PR #80, opened 2021-12-01
Readme.md Rename from Open Source CSS Variables -> Props

alex-page:


argyleink:
i go back and forth all the time on whether to pitch them as props or vars.. tough cuz there's an official name and then the street name. thanks for the offer here, but i'll stick with it as it currently is!

PR #92, opened 2021-12-03
add easing

airen:
Add some easing


argyleink:
I totally had these originally and removed them! https://github.com/argyleink/open-props/commit/48c4901e61ab6154e412d6f43a5cc3866c83c5ab
- i never really liked any of them... they kinda stink right?
- they're really hard to remember, the naming stinks right? 🙂 
- they're also available as postcss plugins n stuff already

thoughts?

airen:
> I totally had these originally and removed them! [48c4901](https://github.com/argyleink/open-props/commit/48c4901e61ab6154e412d6f43a5cc3866c83c5ab)
> 
> * i never really liked any of them... they kinda stink right?
> * they're really hard to remember, the naming stinks right? 🙂
> * they're also available as postcss plugins n stuff already
> 
> thoughts?

These quadratic Bessel curves are very close to Easing Functions Cheat Sheet (https://easings.net/). Although it is a little difficult to remember, they are practical. It would be too heavy to use the PostCSS plugin alone.

image

argyleink:
the way i tried to help with "difficult to remember" is by aligning limiting the strengths to 5 and just increment the number.. ease-1 being the weakest and ease-5 being the strongest.  so much easier than quad, quart, expo, etc..

i super appreciate the thoughts here, but folks will have to grab those classic easings by hand when using open props.

PR #97, opened 2021-12-04
Import `JSON`

### LINKED ISSUE #95: Add `json` to CDN and NPM as a way to consume the props
see https://github.com/argyleink/open-props/discussions/87

it'll be a limited set of values and wont be adaptive by default, but folks should be able to fold individual sets of props into their token systems this way.

argyleink:
fixes #95, cc @roblevintennis and #87 

this PR/branch is a POC for building JSON files alongside the css files. there's still lots to do, but this kicks it off.

todo:
- [x] fix trailing commas
- [x] output files to project root not src
- [x] add files to the exports in package.json
- [x] update docsite with info about json option
- [x] match a design system json schema
- [x] test


argyleink:
Screen Shot 2021-12-03 at 9 18 33 PM

added a place to specify which props to be turned into JSON

src/_create-props.js
```diff
@@ -45,6 +45,33 @@ const individual_colors = {
...
+
+const buildPropsJSON = ({filename,props}) => {
+  const type = filename.split('.')[1]
+  const file = fs.createWriteStream(`${type}.json`)
+
+  file.write(`{\n`)
+  file.write(`  "${type}": {\n`)
+
+  const entries = Object.entries(props).map(([prop, val]) => {
+    return `    "${prop}": { "value": "${val}" }`
```
roblevintennis, line 66: Adding a loop to generate JSON like this definitely is an option, but I'm wondering how well this scales 🤔 

If you look at the discussion (I know it got big lol) I felt that after discovering style dictionary and other similar tools, you'd want to let JSON be the source. But, you likely have motives for keeping that in JavaScript I don't yet understand.

I wanted to just see how SD worked, so I made a little test and created blue and cyan JSON source files based off your key/vals and configured SD to generate both individual files e.g. blue.css, cyan.css, and a color.css (since I noticed you have a combined one that's generally preferred in OP):

image

image

The reason I personally like this, is it's declarative with no code basically leveraging SD. Style Dictionary's whole reason for being is to take these JSON files and output to various formats and more and more get included, its robust, etc. Further, having the source in JSON makes it more likely that beginners can contribute e.g. you and I see what's here as a simple loop of `Object.entries`, some massaging of a string, and then writing stuff out via nodejs...but many will be intimidated.

All to say, for OP maybe all this is totally fine for now and this is certainly a viable solution! I may try to convert my own AgnosticUI's hard coded custom properties to JSON tokens right now. I'll send a link back in case you're interested (but I suspect you have good reasons to keep the source in JavaScript as I mentioned 😋  so I sense you'll want to stick with the hand rolled library-free approach you have here).

Thanks for being open to the idea of design tokens at all in the first place Adam!

roblevintennis: Sorry to belabor but I thought of probably a more compelling reason the have the source in JSON. The whole bi-directionality thing. If eventually you reach the holy grail of stuff being importable into FIgma et al, wouldn't it be nice if, in turn, the design makes some changes to those, hits export, and then THAT generated JSON becomes the new updated source of truth. I don't believe the strategy of taking the JavaScript source files and exporting to JSON is going to jive with that and what's happening is the developers maintain all control of what's "the real" truth if that makes sense.

For me, this is why I'd shoot for having the JSON, standards based source over JavaScript source.

argyleink: JSON as a compile target allows other people to run it through transformers. OP has more to offer than what json based tokens can describe. I eventually want to move back to a CSS as source, since the web as a primary target is very important.

How come OP outputting JSON isn't good enough? Then folks take the JSON and do as they need? Can be the source of truth to them and their needs? 

The use case here is so folks can use "some" OP props, to transform them into tokens for use across their platforms and tools yeah?

roblevintennis: First off I want to again graciously thank you for the library and effort in the first place Adam.

> How come OP outputting JSON isn't good enough?

Ah well maybe it is good enough. I think I've mentioned the directionality potential if you can derive from other tooling and bring it back. But, I certainly understand when things start to go too far away from intended scope. Ultimately, these are all just suggestions and ideas ;-)

> The use case here is so folks can use "some" OP props, to transform them into tokens for use across their platforms and tools yeah?

Again, this is definitely a great step in the right direction and if exporting from OP to JSON tokens one-way is the goal, it's certainly going to be very very helpful. I do think there are several compelling use cases around design tokens. It's fine if OP's intended use is to be the source of truth. It's not what I had pictured as an ideal, but who says I know what's best 😄 

___

I can certainly relate to the need for limiting scope on an open source project that you're doing out of love in your spare time. For AgnosticUI, I saw another library have this really nice flexible CSS props API and I immediately wished I had thought of it (Vuetensils if you must know haha). First thought was I gotta support this flexible API too. But, I realized it's more important to stay true to your original vision and be realistic about how much time and effort you have to make that vision a reality. So I shelved it. I probably won't even do it. So yeah, I can relate Adam — definitely, I absolutely support that you have a vision for OP and you should absolutely stick with it! Apologies if my ideas were coming off too imperative or demanding. Certainly didn't mean to do that.

argyleink: i feel like you're assuming i'm too busy to make things "bidirectional" by going with JSON as the source or I dont understand the value or i want to keep scope low, but i keep feeling your missing how different OP is from design tokens, and missing how JSON formatted to these design token standards would severely limit OP on the web in CSS. providing JSON (and even the provided JS) is a downgraded experience, less can travel and package that way. we can supply these limited JSON tokens to folks who can do whatever transformations they like. then OP doesnt need to balance a bunch of transformations.

i appreciate you reiterating the value of JSON as source, and showing me these tools, but i think you should peek at everything OP has to offer, not just the colors or sizes that have direct crossover with the mainstream design token world. you're not wrong in your views or the way you roll your build system and components + tokens, but i think OP can fully unblock folks who live in these design token JSON worlds, by just supplying the JSON where applicable. that's all a consumer is looking for, some JSON source, they can take it from there. 

Open Props will likely move back to CSS as a source at some point, and JS and JSON will be compile targets, each with their own types of downgrades. **The reason is these are web variables first, design tokens 2nd**. but, of course, this is subject to change, i'm not 100% close minded on it.

roblevintennis: I feel you and you're probably right I should get to know open props a bit better to understand the parts that don't translate to JSON well, the full value proposition,etc. I'll play with it more this week to better understand that 👍🏽

argyleink:
`open-props.tokens.json` meet the [design token standard](https://design-tokens.github.io/community-group/format/), though could be improved by adding types. 

will be available via npm as `require('open-props/json')` or via cdn as `https://unpkg.com/open-props/open-props.tokens.json`

roblevintennis:
Noice! Thanks for adding design token support @argyleink 🎉

PR #100, opened 2021-12-04
Linearly Scaled UI

airen:
Tutorial: https://css-tricks.com/how-to-get-a-pixel-perfect-linearly-scaled-ui/ by @Georgi Nikoloff 
 Demo: https://codepen.io/airen/full/oNWwdoZ by @Airen
 

Dynamically scale CSS values based on viewport width. Use CSS custom properties, comparison functions to build a linearly scaled UI to achieve a pixel perfect restored UI.


argyleink:
interesting technique, i could see that working well for charts and parts of a UI for sure. 

i don't think it's a good fit for open props though. 
i'm going to close for now, but keep this archived in case it comes up again. thanks for the recommendation!

PR #126, opened 2021-12-14
Masks

argyleink:
zigzags, scoops, scallops and corner cuts

discussion in #71 

todo:
- [x] add 3 angle cuts
- [x] add 3 square cuts
- [x] update docsite with example section
- [x] update docsite with usage section
- [x] add tests
- [x] fix scalloped vertical
- [x] verify bundled file is as minimal as possibble
- [x] add credit to Temani in src and docsite

Screen Shot 2022-08-04 at 9 51 35 PM


Afif13:
I think if the mask are going to be static we can still optimize them. My initial idea was to have the same code for all and only adjust the angle as a variable but we can still have an optimized version for each case.

For the Square cut we can do it with one gradient like below:

```css
--mask: conic-gradient(from -90deg at 1rem 1rem,#0000 90deg,#000 0)  0 0/calc(100%  - 1rem) calc(100%  - 1rem)
``` 

Demo: https://codepen.io/t_afif/pen/OJxpyOo

Afif13:
For the angle one we can do this: https://codepen.io/t_afif/pen/YzrZyaY

Casmaine69 (approved):

argyleink:
nice, something to see here https://open-props--pr126-mask-props-81xijln5.web.app/#masks

argyleink:
need to fix `scalloped vertical`

Screen Shot 2022-08-04 at 9 38 00 PM

Afif13:
You can fix it that way:

```css
--mask-edge-scalloped-vertical: linear-gradient(#000 0 0) center / 100% calc(100% - 2 * 20px) no-repeat, radial-gradient(circle farthest-side,#000 97%,rgba(0,0,0,0)) 50%/ calc(1.9 * 20px) calc(2 * 20px) repeat space
```
I updated 0 to 50% so it becomes centered

Afif13:
by the way, it looks super cool on the website. Can't wait to have more variations! :+1:

Afif13:
I see you made different sizes for the cut-corner which is good. Can we do the same for the edges?
I know we may end with a verbose code if we will have different variations but I was thinking about a variable `--mask-edge-size: 20px` that will replace the `20px` you are setting. 

We can have the variable inside `:where(html) {}`. If anyone want to update it, he can. Otherwise the default 20px will apply.

Afif13:
Also, I just got a better idea for the circular cut

instead of this 

```css
--mask-corner-cut-circles-1: radial-gradient(circle 1rem,#0000 99%,#000) 10000% 10000% / 99.5% 99.5%;
--mask-corner-cut-circles-2: radial-gradient(circle 2rem,#0000 99%,#000) 10000% 10000% / 99.5% 99.5%;
--mask-corner-cut-circles-3: radial-gradient(circle 4rem,#0000 99%,#000) 10000% 10000% / 99.5% 99.5%;
```
We can do this

```css
--mask-corner-cut-circles-1: radial-gradient(1rem at 1rem 1rem,#0000 99%,#000) -1rem -1rem;
--mask-corner-cut-circles-2: radial-gradient(2rem at 2rem 2rem,#0000 99%,#000) -2rem -2rem;
--mask-corner-cut-circles-3: radial-gradient(4rem at 4rem 4rem,#0000 99%,#000) -4rem -4rem;
```

Sorry for the last minute changes 😅 but this last syntax should be better (I just got the idea now) and also give better result. The first one relies on calculations that involve rounding so it can break in very rare case while the last one relies on integers

PR #130, opened 2021-12-17
docs: add autocomplete guidelines for users that use VScode as their primary IDE

Caleb335:
Hi @argyleink,

Kindly review.


argyleink (changes requested):
ty! the core is pretty much all done 🙂 

a few requests i hope are ok!
- can you slim down the commit to just the vscode section, looks like a linter or formatter ran across the whole file
- remove the package-lock from the PR too, i'll make sure that gets an update
- can you remove the header "getting started"
- turn the paragraphs it into a numbered list

I imagine the section could be a paragraph and a few numbered bullets with the code sample, something like this:

> VScode has intellisense which the CSS Var Complete extension can tap into and pass a path to the node_modules folder when NPM's open-props is installed:
> 
> 1. In your VScode environment, navigate to the extensions marketplace, and install the CSS Var Complete extension.
> 2. Install Open Props from NPM
> 3. Add the NPM Open Props CSS file path to your .vscode/settings.json file, like this:
> 
> [great code example showing path to node_modules]

Caleb335:
Awesome! Thank you for the feedback.

I'll start working on them.

Caleb335:
Hi @argyleink, I made the changes you requested. Kindly take a look

argyleink (approved):
rad, thanks for kicking this all off and doing the research!

Caleb335:
> rad, thanks for kicking this all off and doing the research!

Hi @argyleink,

Would it be nice if the autocomplete docs also contains the process of setting it up for other IDEs like atom, Sublime etc.

I can start making my research if you permit me.

argyleink:
sounds great! helping folks get it setup in Atom and Sublime would def be nice 🙂 
thanks for offering more help!

Caleb335:
> sounds great! helping folks get it setup in Atom and Sublime would def be nice 🙂 thanks for offering more help!

Anytime! 😎

Caleb335:
I've started my research about autocomplete in atom, and I've found a use-case already.

Hopefully,  I'll send a PR by the end of the week.

Issue #133, opened 2021-12-20
Normalize `[tabindex]` shouldn't select -1 values


argyleink:
There's a bug with sveltekit this will fix, but it's also a clearer selector based on it's intent.

dkvsky-repo:
I think this would work `[tabindex]:not([tabindex*="-"])`.

argyleink:
fixed in `1.0.14`

PR #134, opened 2021-12-21
Updating tabindex selector to exclude negative values

dkvsky-repo:
Updating tabindex in `normalize.css` to exclude negative values.
Issue #133


argyleink (approved):
nailed it 💯

PR #136, opened 2021-12-22
adding backgrounds to 'why' section icons

sharu725:
Light mode
image

Dark Mode
image


argyleink:
not bad! but i liked the more subtle color use presently used. thanks tho!

Issue #115, opened 2021-12-08
[docs] color swatches in hides color name


shrirambalaji:
As much as I like the animation of the swatches they seem to block the name unfortunately 😥. The other issue I see is that when the drop down is clicked the size of the swatch expands (not sure if that was intended or no)

https://user-images.githubusercontent.com/11358903/145268230-59665553-1170-4a80-b6d3-652a612eaa50.mp4

I'd like to help fix this issue, would love some pointers  @argyleink! Also would like to know if you had considered regular square swatches (like in tailwind docs maybe?) and the naming the colors inside the square or below each square?

argyleink:
Hi Shriram! 

I thought the flexible swatch expansion was neat/fine. Just some responsive components 🙂 ?

Maybe to help legibility of numbers on swatches we add a tooltip instead of an inline number? then your pointer wont be over it? I'd be open to seeing prototypes you make of enhanced ux?! down to PR some thoughts?

shrirambalaji:
That makes sense. I'll try out some changes and we can finalize what looks best!

WebMechanic:
the flexible swatches are fine, but the label variation add visual unrest. Something like this might be helpful:
```css
.open-colors h4 {
    min-width: 6.66ch;
}
```

PR #140, opened 2021-12-30
Update color swatches

shrirambalaji:
Trying to address https://github.com/argyleink/open-props/issues/115

Proposed Changes:
- Fix layout shift by setting colors width to 100%.
- Fix swatch visibility on resize.
- Make swatch rectangular.
- Keep Existing behaviour of showing number on hover.

https://user-images.githubusercontent.com/11358903/147774803-e8b5ddb9-ec3f-438a-bfdd-6c1d2dff2a9a.mp4

@argyleink Let me know your thoughts on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 dist/
 .cache/
 node_modules/
+.vscode/
--- a/docsite/index.css
+++ b/docsite/index.css
@@ -443,7 +443,7 @@ pre[class*="language-"] {
 
   & > h4 {
     counter-reset: basic -1;
-    font-size: var(--fontsize-fluid-1);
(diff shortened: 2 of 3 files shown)


argyleink:
not sure this really tackles the problem you outlined in the issue #115, as that part of the ux is the same? looks like the current PR is mostly style changes and a layout shift fix?

shrirambalaji:
@argyleink I wanted to keep ux changes the same and address the two main problems of layout shift and avoid hiding the colors name when hovered upon.

To completely remove the "number-on-hover" ux we could do either a tooltip or display a row of numbers on the top of the grid similar to `th` elements .

argyleink:
thanks for these ux thoughts, going to close this for now if that's ok

Issue #21, opened 2021-09-21
Figma plugin to add colors and sizes for the designer


argyleink:

hchiam:
Some links that might be helpful:

https://www.figma.com/plugin-docs/setup/

https://help.figma.com/hc/articles/360039958874-Make-plugins-for-the-Figma-Community

hchiam:
there's even a visual UI to create plugins and plugin templates! https://help.figma.com/hc/articles/360042786733

hchiam:
@argyleink, you're thinking of autocomplete, just like in VSCode https://github.com/argyleink/open-props/issues/28, right? so if someone's in figma they can type `var(--pi` and then it'll suggest something like `var(--pink-0)`? (or a list)

Simulated autocomplete:
<img width="456" alt="simulated_autocomplete" src="https://user-images.githubusercontent.com/18131787/147791168-f5eb5ce7-077d-481c-b0c3-3aa045143916.png">

argyleink:
more like importing tokens, like shown here https://jansix.at/resources/figma-tokens. maybe we piggy back on [that plugin](https://jansix.at/resources/figma-tokens) and offer a format it accepts and makes readily available to designers?

hchiam:
oh wow, that's an awesome figma plugin! we can generate open-props tokens into a json file and then paste it into figma tokens JSON tab, or hit the Import button

hchiam:
i could work on creating a script like _create-props.js or pass an option to it, then add a `script` named `"gen:figma"` (or `"gen:tokens"`?)﻿

argyleink:
fixed by #144

PR #144, opened 2022-01-08
generate JSON for figma-tokens plugin to use

hchiam:
for #21

partly set up figma tokens

@argyleink, feel free to suggest changes! (especially for the "other" group - figma-tokens recognizes only certain names for the top-level groupings in the JSON)

I kinda like having the token names the same as the CSS variables

### DIFF
--- a/src/_create-props.js
+++ b/src/_create-props.js
@@ -116,6 +116,11 @@ Object.entries(jsonbundle).map(([key, token]) => {
 const FigmaTokens = fs.createWriteStream('../open-props.figma-tokens.json')
 FigmaTokens.end(JSON.stringify(figmatokens, null, 2))
 
+const figmatokensSYNC = { 'open-props': { ...figmatokens } }
+
+const FigmaTokensSync = fs.createWriteStream('../open-props.figma-tokens.sync.json')
+FigmaTokensSync.end(JSON.stringify(figmatokensSYNC, null, 2))
+
 const buildPropsStylesheet = ({filename, props}) => {
   const file = fs.createWriteStream(filename)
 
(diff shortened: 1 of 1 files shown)


hchiam (commented):
Here's a quick look at what the generated JSON looks like: 
image

hchiam:
and then once the .json file is deployed to unpkg, a designer can copy a link into the Sync tab (instead of manually copy-pasting all the contents of a JSON file)

argyleink:
rad work! def be sweet when this is just an unpkg url away! 

i have an issue tho: i cant get my plugin to be as organized as yours!? tell me your secrets!

here's my URL (https://jsonbin.io/61d91d632362237a3a343e96) and then a preview of it syncing with the plugin 

Screen Shot 2022-01-07 at 9 38 16 PM

the JSON seems fine to me, looks like you followed their docs super well! this a user (me) error? the ux of this plugin is a bit.. tough in some ways. how did you get all your props to show up under the group "open-props"? is it because i'm using the url feature and not jsonbin?

Screen Shot 2022-01-07 at 9 45 10 PM

the PR otherwise looks solid! colors look great in there 🙂

hchiam:
oh, i hadn't tried the Sync tab yet, i manually pasted into the JSON text box. turns out we need to wrap all the tokens in one more key where that lone top-level key could be "open-props", like where the `"global"` is in this example: https://raw.githubusercontent.com/six7/figma-tokens-example/main/tokens.json

commit forthcoming

hchiam (commented):
turns out you need a top-level name for the Sync tab to work (and no top-level name for the JSON tab) https://github.com/hchiam/figma-tokens-sync-example

argyleink (approved):
thanks, this looks great! 
will work on docs next and publishing so the url goes live 👍🏻

argyleink:
boom https://unpkg.com/open-props/open-props.figma-tokens.sync.json

Issue #139, opened 2021-12-30
Inner shadows mislabeled


brittneypostma:
Inner shadows are labeled out of order, should be (0-4), instead we have 0 then jumps to 7-10.
![image](https://user-images.githubusercontent.com/45889730/147768719-d85b96ad-a97c-4038-8c01-8b4871c8a560.png)

argyleink:
oh! dang css counters look like they reset to 0 then resume(?) at 7 lol, i'll check it out!

argyleink:
oh, which browser?

argyleink:
looks like only Firefox reproduces this bug.. might be a CSS counters bug in Firefox?

brittneypostma:
Sorry, just seeing these, yes I was in FireFox Dev, that is interesting.

argyleink:
fixed by #145

PR #145, opened 2022-01-09
fix: shadow counter on firefox

equinusocio:
Ref. https://github.com/argyleink/open-props/issues/139


argyleink (approved):
very nice, improvements all around 🙂

PR #172, opened 2022-02-01
remove vertical-align as it doesn't apply on display: block-elements

benbender:
f.e. mdn says on https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align:
`Note that vertical-align only applies to inline, inline-block and table-cell elements: you can't use it to vertically align block-level elements.`

### DIFF
--- a/src/extra/normalize.css
+++ b/src/extra/normalize.css
@@ -99,7 +99,6 @@
 
 :where(img, svg, video, canvas, audio, iframe, embed, object) {
   display: block;
-  vertical-align: middle;
 }
 
 :where(img, svg, video) {


argyleink:
a lot of those elements are `inline-block` though, maybe this the PR can remove the block level elements from the selector list instead of remove the style?

benbender:
Hmm. Real question would be if those elements are "normalized" by default in all of the targeted browsers of open-props. If yes, I would agree that the `display: block` can be removed. If not, I would keep it as a solid baseline to build upon, because otherwise there could be subtle bugs. I'm unsure...

PR #176, opened 2022-02-01
docs: update colors-hsl usage 

x-dune:
Update colors-hsl usage in the **Modify Opacity Example**.

Also, it looks like `colors-hsl` is not included in `open-props.min.css`. Not sure if that's intentional.


argyleink (approved):
nice, thanks for fixing my bad doc update!
`colors-hsl` is a separate import yeah

PR #177, opened 2022-02-01
DOCS: Fix counters in Firefox

kevin-powell:
Firefox doesn't properly support `counter-reset` if the parent also uses counter-reset for the same counter.
(oddly enough, it does reset it for that individual element, but not the other elements using the same counter).

This has an impact on the colors section, which resets the counter on the `h4`s. When you hover on them in FF, because the reset wasn't carrying over, the numbering goes from 1 to 10, rather than 0 to 9.

open-prop-wrong-numbers

Firefox would work with a `counter-set`, rather than `counter-reset` on the parent, but atm it's not supported in Safari, so this is a bit of a workaround using a pseudo-element on `.count-em` to reset it instead since Firefox is fine with different siblings resetting the same counter.

This `::before` caused some layout issues anywhere `count-em` was used to show layout classes, such as sizes, which is why the `position: absolute` has also been added.

You can see all three approaches in this [very simple example in this codepen](https://codepen.io/kevinpowell/pen/bGYpwgX), with the first example not properly resetting if viewed in Firefox.

### DIFF
--- a/docsite/index.css
+++ b/docsite/index.css
@@ -1215,7 +1215,7 @@ pre[class*="language-"] {
 }
 
 .count-em {
-  counter-reset: basic;
+  counter-set: basic;
 
   & > *::before {
     counter-increment: basic;
--- a/docsite/index.css
+++ b/docsite/index.css
@@ -1218,7 +1218,9 @@ pre[class*="language-"] {
 
   &::before {
   content: '';
-  counter-reset: basic;}
+  counter-reset: basic;
+  position: absolute;
+}
 
   & > *::before {
     counter-increment: basic;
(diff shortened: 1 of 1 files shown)


argyleink (approved):
verified fixed, thank you!

myfonj:
Historical note: for some period of time, spec dictated the behaviour Firefox exhibited, but it was reverted after some time. Yet even now we are left in a quite confusing state where `counter-reset` has a meaning more like "`~counter-init-new-scope`", and `counter-set` is more like "`~counter-assign-deepest-scope`".

Some context and remarks in https://bugzilla.mozilla.org/show_bug.cgi?id=1729498

PR #187, opened 2022-02-11
add `src/` folder to package.json exports

### LINKED ISSUE #186: Add the `./src/` folder to exports
At present, the only way to use Open Props in a `"type": "module"` setup it is through the `"module": "open-props.module.js"` file which slightly impairs the ability of bundlers to optimize the package since it's operating on a pre-combiled bundled. It would be helpful if setups with bundlers could directly import the original `./src/index.js`. 

Furthermore, I'd like to also to import the individual files into js e.g `import { Teal } from "./src/props.colors.js"` and have my bundler take care of the treeshaking etc. rather than import the whole `open-props.module.js`. 

This is a feature request to export the `src`folder:

```diff
"exports": {
    ".": {
      "import": "./dist/open-props.module.js",
      "require": "./dist/open-props.cjs",
      "default": "./dist/open-props.cjs"
    },
+   "./src": "./src",
	"./style": "./open-props.min.css",
```

### Additional Context
Some bundlers (e.g Vite) strictly only use the `exports` field if it exists. That is, acessing the `./src` folder throws '`./src` is not exports'.

kwangure:
Fixes #186.


src/index.js
```diff
@@ -11,7 +11,7 @@ import {StaticShadows} from './props.shadows.js'
 import SVG from './props.svg.js'
 import Zindex from './props.zindex.js'
 
-export default OpenProps = {
+export default {
```
argyleink, line 14: this change a suggestion because it wasnt doing anything meaningful? or you needed to change it to fix something?

package.json
```diff
@@ -26,6 +26,15 @@
...
     },
+    "./src": "./src/index.js",
+    "./src/props.animations.js": "./src/props.animations.js",
+    "./src/props.aspects.js": "./src/props.aspects.js",
+    "./src/props.borders.js": "./src/props.borders.js",
+    "./src/props.colors.js": "./src/props.colors.js",
+    "./src/props.colors-hsl.js": "./src/props.colors-hsl.js",
+    "./src/props.easing.js": "./src/props.easing.js",
+    "./src/props.fonts.js": "./src/props.fonts.js",
+    "./src/props.gradients.js": "./src/props.gradients.js",
```
argyleink, line 37: missing sizes, shadows, and maybe something else? 
(i didnt do a full scrub)

argyleink (changes requested):
thanks for taking the time to make a PR!

package.json
kwangure: Yes. I stopped scrolling too early.

src/index.js
kwangure: Fix. It's an assignment to a non-existent variable. The alternative is:

```javascript
const OpenProps = { ... };
export default OpenProps;
```

argyleink (approved):

PR #191, opened 2022-02-15
@parcel/css-cli to replace postcss build commands

argyleink:
todo
- [ ] move the poc command line added to package.json to all the postcss commands
- [ ] test before and after filesizes (for funsies)
- [ ] test and inspect all output to ensure no regressions
- [ ] determine how to, or if OP continues to, combine selectors

this proves the groundwork though, and given the output is smaller, than this PR is a go 👍🏻

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "gen:op": "cd build && node props.js '' true",
     "gen:nowhere": "cd build && node props.js '' false",
     "gen:prefixed": "cd build && node props.js 'op'",
+    "parcel": "parcel-css --bundle --custom-media --nesting --minify -t 'last 2 versions' src/index.css -o docsite/parcel.css",
     "lib:js": "npm run bundle:pre-edit && microbundle --no-sourcemap && npm run bundle:post-edit",
     "lib:all": "postcss src/index.css -o open-props.min.css",
     "lib:normalize": "postcss src/extra/normalize.css -o normalize.min.css",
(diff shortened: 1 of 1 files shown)


argyleink:
much faster builds on this 2010 machine 👍🏻

devongovett:
Hey @argyleink! Anything you need help with to get this in? 😄

argyleink:
i dont think there's any technical stuff in the way, it's me mostly ironing it all out and doing all the cross checks to ensure to regressions 🙂 

thanks for checkin in!

argyleink:
just spent an hour doing builds and testing/inspecting output! there's a lot of great things going on in this, speed being a huge one. 

having the logical properties syntax lowered, even place-items, are adding lots of lines of code. makes me want to bundle 2 versions and do the like, modern css vs syntax lowered css builds. if i offer them on the cdn or via exports, they could pick which they snag. i don't see any ways of specifying via cli options to not lower logical props? feature request 🙂 btw though, the syntax lowering of logical props is super impressive and rad.

i love the syntax lowering of color! love the clean output.

it looks like i could switch to parcel for just builds of the basic files right away, but not normalize since it almost doubled from logical property lowering. but, it's so fast. the main consumed files are mostly values, not logical properties, and it's really only a few lines of change to package.json.

i dont think i'd want to juggle both postcss and a parcel tool, but maybe i do! will think and check more.

devongovett:
Logical properties lowering is based on the targets specified to the CLI. You have "last 2 versions", which would include eg ie 10 and 11. Usually it's better to do a browserslist query based on market share percentage. You can play around here: https://browserslist.dev

argyleink:
todo: try with the latest Lightning release

bjorntheart:
@argyleink is this still something you'd want for OP? If so, my [PR](https://github.com/argyleink/open-props/pull/474) you recently merged would need an update to use lightningcss. I'm happy to make those changes once/if this is merged.

argyleink:
> @argyleink is this still something you'd want for OP? If so, my [PR](https://github.com/argyleink/open-props/pull/474) you recently merged would need an update to use lightningcss. I'm happy to make those changes once/if this is merged.

i've been testing it more in OPv2 branch, and still not entirely sure I'll go with it in the end. but thanks for checking and the offer! for now I'd say hold off 🙂

PR #193, opened 2022-02-17
Fix build on windows devices

Que-tin:
Fix build on windows devices by manually setting undefined for a not defined prefix.

### DIFF
--- a/build/props.js
+++ b/build/props.js
@@ -18,7 +18,14 @@ import {toJSON} from './to-json.js'
 import {toTokens} from './to-tokens.js'
 import {toFigmaTokens} from './to-figmatokens.js'
 
-const [,,prefix,useWhere] = process.argv
+let prefix = process.argv[2];
+const useWhere = process.argv[3]
+
+// On windows devices the prefixs is "''" if not set
+if(prefix === "''") {
+  prefix = undefined;
+}
+
 const selector = useWhere === 'true' ? ':where(html)' : 'html'
 
 const mainbundle = {
--- a/build/props.js
+++ b/build/props.js
@@ -18,13 +18,7 @@ import {toJSON} from './to-json.js'
 import {toTokens} from './to-tokens.js'
 import {toFigmaTokens} from './to-figmatokens.js'
 
-let prefix = process.argv[2];
-const useWhere = process.argv[3]
-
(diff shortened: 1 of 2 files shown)


build/props.js
```diff
@@ -18,7 +18,14 @@ import {toJSON} from './to-json.js'
...
 
-const [,,prefix,useWhere] = process.argv
+let prefix = process.argv[2];
+const useWhere = process.argv[3]
+
+// On windows devices the prefixs is "''" if not set
+if(prefix === "''") {
+  prefix = undefined;
+}
+
```
argyleink, line 28:
```suggestion
const [,,prefix='',useWhere] = process.argv
```

then in [`to-stylesheet.js`](https://github.com/argyleink/open-props/blob/main/build/to-stylesheet.js#L17) update that prefix if statement to check for empty strings too? normalizes empty string as the default.

argyleink (changes requested):

build/props.js
Que-tin: Did some changes, didn't quite get what you were trying to tell me., but I realized, that there is already a check for this in [to-stylesheet.js](https://github.com/argyleink/open-props/blob/main/build/to-stylesheet.js#L17) as you said. Thats why I just added the check in there, does the same, it is much cleaner.

build/to-stylesheet.js
```diff
@@ -14,7 +14,7 @@ export const buildPropsStylesheet = ({filename,props}, {selector,prefix}) => {
     if (prop.includes('-@'))
       return
 
-    if (prefix)
+    if (prefix && prefix !== "''")
```
argyleink, line 17:
```suggestion
    if (prefix && prefix !== '')
```

Screen Shot 2022-02-23 at 9 20 22 PM

curious why you are checking for quotes instead of empty string?

Que-tin: Because that's what seems to be passed in as arg from the terminal.

While Unix systems pass in undefined for empty args for some reason Windows passes in '' and that then is interpreted as string so you get an string with quotes as an result.

Tried everything else that's the only solution that works for me. I really need to check for that.

Maybe wait until you get your windows machine back before merging and check yourself. I am on Windows 10 Node Version 16.14

PR #196, opened 2022-02-23
fix prefers-contrast to draft spec

philstenning:
Found that the prefers-contrast prop in the media queries look like it should be: more | less not high | low
checked the w3c spec and MDN

https://drafts.csswg.org/mediaqueries-5/Overview.bs

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast 

and a demo from a link in discord https://codepen.io/argyleink/pen/RwgzJXV


argyleink (approved):
awesome, thank you!

PR #198, opened 2022-02-24
Update Dokumentation

chimp1nski:
**Update gradient import statement:**
`import Gradients from 'open-props/src/gradients';`

**Remove all sorts of whitespace**

### DIFF
--- a/docsite/index.html
+++ b/docsite/index.html
@@ -884,7 +884,7 @@
                     import {StaticShadows} from 'open-props/src/shadows';
 
                     // import the gradients
-                    import Gradients from 'open-props/src/colors';
+                    import Gradients from 'open-props/src/gradients';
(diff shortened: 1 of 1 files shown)


chimp1nski:
I was kind of confused by the Gradient import since to me it looks like a "default" import statement - which would mean that the Colors get imported and just named "Gradient".
Looks to me like this could be a copy paste error?

argyleink (changes requested):
nice catch thanks! looks like a copy paste error yeah

can you reduce the PR to just the 1 line change? whitespace is managed in production builds.

chimp1nski:
Hey, I updated the PR to only include the import change.

Took me a while to find my editor settings to not automatically trim the whitepsace xD

Issue #199, opened 2022-02-24
Imports not working in TypeScript project.


chimp1nski:
When I am trying to import any of the Props into a TypeScript project,

`import sizes from "open-props/src/sizes"`

 I get a ts(2307) Error: _Cannot find module 'open-props/src/sizes' or its corresponding type declarations._

---

Svelte-kit project,
TypeScript version: 4.4.3,
target: "es2020",

---

A workaround I found for this is declaring my own modules for the exports[] in package.json.

open-props.d.ts
``` typescript
declare module "open-props/src/sizes";
```

---

I am not sure if this is a problem with OpenProps or TypeScript / SvelteKit.
Once I find time, I will try and set it up in a React Project and see what happens there.

argyleink:
`import sizes from "open-props/src/props.sizes.js"`

does that resolve the issue? this way you're referencing a file directly and not leaning on the [exports map in package.json](https://github.com/argyleink/open-props/blob/main/package.json#L39). 

not sure why declaring the module resolves it tho? the file you're importing is some [basic ecma js](https://github.com/argyleink/open-props/blob/main/src/props.sizes.js) 

thanks for reporting, we'll get ya through it, and maybe i can update the docs to help others in the future

chimp1nski:
Unfortunately importing via `open-props/src/props.sizes.js` doesn't work.
I forgot to mention that.

I am getting a different error though:
`[vite] Internal server error: Missing "./src/props.sizes.js" export in "open-props" package`

It's kinda odd that the vite error is refering to "./src".
Might actually be a bug in the svelte-kit vite configuration.

It actually doesn't really block me since I am not actually using the props in javascript.
Was just trying things out.

argyleink:
glad you're not blocked, but still a bit disconcerting. are you able to import other props, or all props kinda give you the same issues with missing exports?

chimp1nski:
Sizes was just used as example here. I am actually not able to import any props.

I was digging a bit deeper and from what I can see, the index.js in /src is re-exporting the props.xyz.js objects through 

```javascript
const OpenProps = mapToObjectNotation({
  ...Animations,
  ...Sizes,
  ...Colors,
  ...ColorsHSL,
  ...Fonts,
  ...Borders,
  ...Aspects,
  ...Easings,
  ...SVG,
  ...Gradients,
  ...StaticShadows,
  ...Zindex,
});

```

[...]

PR #223, opened 2022-04-10
typings setup

hchiam:
I set up the [tsconfig.json](https://github.com/argyleink/open-props/compare/main...hchiam:typings-setup-issue-199?expand=1#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0) so that all the declaration files (.d.ts) in the /types folder are automatically generated from the JS files, except for the index.d.ts declaration file (which contains `OpenProps`). The index.d.ts file is simply copied from /src to /types, and the file will need manual updates if other /src files are added. 

All of the above is done by this added package.json command:

```sh
npm run gen:types
```

So now all props autocomplete on the top-level `OpenProps` object! :)

image


hchiam:
For #199

types/index.d.ts
```diff
@@ -0,0 +1,28 @@
+import propsAnimations from '../types/props.animations'
```
hchiam, line 1: All the other .d.ts files are auto-generated.

hchiam (commented):
Ready for initial review. I plan to use the following repo to do more testing: 
https://github.com/hchiam/svelte-kit-typescript-open-props

hchiam:
This PR #223 seems to fix the ts(2307) Error mentioned in issue #199 and the issue with typing `OpenProps`, as tested in my test repo: 

https://github.com/hchiam/svelte-kit-typescript-open-props/commit/1f8a3612d212b534899a17b610e7e6a66af97ed1 

So instead of `import sizes from "open-props/src/sizes"` or `import sizes from "open-props/src/props.sizes.js"`, this seemed to work: 
```js
import sizes from 'open-props/src'
```

argyleink (changes requested):
Back from vacation, thanks for this work and waiting patiently!
thanks for making the test repo too, very helpful for testing this PR 👍🏻 

I can see a foundation here for TS, but it's missing / not matching the JS API, which I think it should fully support so I don't  need to manage a split in the API in the documentation. 

1) I noticed that the object notation accessors arent auto completed, just the array accessor style. So `import op from 'open-props/src'` and `op['--size-1']` works, but `op.size1` doesnt. the ES module [does that work here](https://github.com/argyleink/open-props/blob/main/src/index.js#L14-L27), probably can add the same logic to `index.d.ts`?
2) `import op from 'open-props/src'` and `import sizes from 'open-props/src'` are the same thing, which is expected. but `import {sizes} from 'open-props/src'` doesnt work unless I add `export const sizes = propsSizes;` to `index.d.ts`. Destructured imports are really nice and should continue to work in TS.
3) Confusing that TS needs `from 'open-props/src'` as it's the same file reference as `from 'open-props'`. We need to teach TS to not require the `/src` in order for it to find the modules.
4) should we add `"types": "./types/index.d.ts"` to package.json?
5) `import sizes from 'open-props/sizes'` should work and it doesnt yet

Ultimately I'm looking to have this example code block work in TS or JS, and the developer shouldnt have to adjust anything

```js
// index.js loading JS object
import OpenProps from 'open-props'; // module
import OpenProps from 'open-props/src'; // unbundled ES module
import Colors from 'open-props/src/colors';

// object notation access is special to OpenProps
console.info(OpenProps.size1);
console.info(OpenProps['--size-1']);

console.info(Colors['--indigo-5']);
```

Those are my 30m review thoughts! All feels like solvable stuff yeah? 
Thanks again, it's a great start!

argyleink:
also worth checking out this PR https://github.com/argyleink/open-props/pull/225 which is seeking to solve the same issue and has an alternative solution that resolves some of my comments above.

hchiam:
this PR is still a WIP - TODO: figure out why both array and dot notations are working in autocomplete.test.ts but not in the test svelte repo

hchiam:
https://github.com/argyleink/open-props/pull/223/commits/31aa59192bfbdae2ac22ab4eb7e0161a0c1aba83 A thing to consider is that JS props can't end with the `@` symbol, so the `microbundle` approach in PR #225 will need to be modified to not auto-generate `.animationFadeIn@`.

argyleink:
> A thing to consider is that JS props can't end with the `@` symbol, so the `microbundle` approach in PR #225 will need to be modified to auto-generate `.animationFadeInAt` instead of `.animationFadeIn@`?

dang, so [jit-props](https://github.com/GoogleChromeLabs/postcss-jit-props#:~:text=in%20.5s%20ease%27%2C-,%27%2D%2Dfade%2Din%2D%40%27%3A,-%27%40keyframes%20fade%2Din) would need updated too, with some sort of breaking syntax change too. it relies on the syntax ending with an `@` to find keyframes for animation props. 🤔  def don't want to break 2 packages.

aaronccasanova:
> A thing to consider is that JS props can't end with the `@` symbol, so the `microbundle` approach in PR https://github.com/argyleink/open-props/pull/225 will need to be modified to auto-generate `.animationFadeInAt` instead of `.animationFadeIn@`?

Property names can include `@` symbols, they just have to be accessed with bracket notation:

Screen Shot 2022-04-30 at 10 12 54 AM

I'm not familiar with `jit-props`, but I imagine if it uses computed member access (e.g. brack notation) there should be no need to introduce a breaking change.

hchiam:
Yes, technically JS property names can have `@` in them if you access them as strings with the bracket notation, but when I did some testing in VSCode, property names with `@` in them may not show up in the autocomplete after you hit `.` and start typing, unless you intentionally try to access them with the bracket notation. I guess maybe not a huge deal-breaker as it's still technically accessible to users. So I'm thinking of reverting the code in this PR that remaps `@` to `'At'`

hchiam:
> Yes, technically JS property names can have `@` in them if you access them as strings with the bracket notation, but when I did some testing in VSCode, property names with `@` in them may not show up in the autocomplete after you hit `.` and start typing, unless you intentionally try to access them with the bracket notation. I guess maybe not a huge deal-breaker as it's still technically accessible to users. So I'm thinking of reverting the code in this PR that remaps `@` to `'At'`

Here's the repo I used to test autocomplete and typings for this PR: 

https://github.com/hchiam/svelte-kit-typescript-open-props/blob/main/my-app/package.json#L37 (WIP)

argyleink:
available in `v1.5.4`!

PR #224, opened 2022-04-15
Add .github and .firebaserc to .npmignore

nucliweb:
## In this Pull Request

Added the folder `.github` ant the `.firebaserc` to th `npmignore` file to avoid to publish in the npm package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,4 @@
+.firebaserc
+.github
 docsite/
 test/


nucliweb:
Btw, I'm thinking that the folder `src` is not used to import the tokens, so really it is not necessary either

argyleink:
> Btw, I'm thinking that the folder src and the file postcss.config.js are not used to import the props, so really it are not necessary either

`postcss.config.js` could be removed yeah, but `src/` should stay so folks can unpkg reference src files if they want

argyleink (approved):

PR #225, opened 2022-04-15
feat: Expose TypeScript declarations

aaronccasanova:
**Do not merge:** See `TODO` section below

This PR updates the `lib:js` build to additionally output TypeScript declarations. [Microbundle](https://www.npmjs.com/package/microbundle) already handles TypeScript projects with zero config and thus only a few tweaks were needed to emit declaration files:
- Added a `tsconfig.json` defining the `dist/types` output directory
- Updated the `package.json` `"types":` and `"exports.types":` keys to resolve the emitted `dist/types/index.d.ts`
- Added the [type-fest](https://github.com/sindresorhus/type-fest) (TypeScript utility types package) as a dev dependency to address the `any` types in the `src/index.js` default export
- Added a `JSDoc` annotation to help `tsc` understand the camel case keys transform on the default export

#### After running `npm run lib:js`
image

#### Example `npm link`ing and using the intellisense in a `.mjs` file
image

### Important

There is still one issue I haven't resolved and that is that the declarations in `.cjs` modules want to resolve values off `OpenProps.default`:
image

TODO:
- [ ] - Debug why `.cjs` files think props are accessible from `OpenProps.default`
- [ ] - Take a second pass at the `tsconfig` `compilerOptions` and ensure the appropriate defaults have been set

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,8 +11,10 @@ import Shadows from './props.shadows.js'
 import SVG from './props.svg.js'
 import Zindex from './props.zindex.js'
 
-const camelCase = (text) => {
-  text = text.replace(/[-]+(.)?/g, (_, c) => (c ? c.toUpperCase() : ''))
+const camelize = text => {
+  text = text.replace(/[-]+(.)?/g, (_, c) => c 
+    ? c.toUpperCase() 
+    : '')
   return text.substr(0, 1).toLowerCase() + text.substr(1)
 }
 
@@ -22,7 +24,8 @@ const camelCase = (text) => {
  * @returns {import("type-fest").CamelCasedPropertiesDeep<T>}
  */
 const keysToCamelCase = (props) => {
-  for (var prop in props) props[camelCase(prop)] = props[prop]
+  for (var prop in props)
+    props[camelize(prop)] = props[prop]
   return props
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -21,7 +21,7 @@ const camelize = text => {
 /**
  * @template T
  * @param {T} props
- * @returns {import("type-fest").CamelCasedPropertiesDeep<T>}
+ * @returns {T & import("type-fest").CamelCasedPropertiesDeep<T>}
  */
 const keysToCamelCase = (props) => {
   for (var prop in props)
--- a/src/index.js
+++ b/src/index.js
@@ -11,20 +11,22 @@ import Shadows from './props.shadows.js'
 import SVG from './props.svg.js'
 import Zindex from './props.zindex.js'
 
-const camelize = text => {
-  text = text.replace(/[-]+(.)?/g, (_, c) => c 
-    ? c.toUpperCase() 
-    : '')
+const camelCase = (text) => {
+  text = text.replace(/[-]+(.)?/g, (_, c) => (c ? c.toUpperCase() : ''))
   return text.substr(0, 1).toLowerCase() + text.substr(1)
 }
 
-const mapToObjectNotation = props => {
-  for (var prop in props)
-    props[camelize(prop)] = props[prop]
+/**
+ * @template T
+ * @param {T} props
+ * @returns {import("type-fest").CamelCasedPropertiesDeep<T>}
+ */
+const keysToCamelCase = (props) => {
+  for (var prop in props) props[camelCase(prop)] = props[prop]
   return props
 }
 
-const OpenProps = mapToObjectNotation({
+const OpenProps = keysToCamelCase({
   ...Animations,
   ...Sizes,
   ...Colors,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "open-props",
-  "version": "1.3.12",
+  "version": "1.3.16",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "open-props",
-      "version": "1.3.12",
+      "version": "1.3.16",
       "license": "MIT",
       "devDependencies": {
         "ava": "^3.15.0",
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -11,20 +11,22 @@ import Shadows from './props.shadows.js'
...
+ * @param {T} props
+ * @returns {import("type-fest").CamelCasedPropertiesDeep<T>}
+ */
+const keysToCamelCase = (props) => {
+  for (var prop in props) props[camelCase(prop)] = props[prop]
   return props
 }
 
-const OpenProps = mapToObjectNotation({
+const OpenProps = keysToCamelCase({
```
aaronccasanova, line 29: https://user-images.githubusercontent.com/32409546/163602878-e20c148e-e9c4-429c-9e69-bd1caab110c9.mov

aaronccasanova:
@argyleink Regarding the TypeScript issue in `CommonJs` modules. As shown below, TypeScript currently thinks tokens are accessible from `OpenProps.default[token]` (on the `default export`). Whereas, they are actually accessible directly from `OpenProps` (e.g. `OpenProps.animationBlink`):

image

I'm struggling to identify the exact issue, but have narrowed things down a bit. Microbundle by default sets the [rollup.OutputOptions.exports to 'auto'](https://github.com/developit/microbundle/blob/fb0a43780a7462f4876955c3412638e51a7adb97/src/index.js#L682) which according to the [output.exports section of the Rollup docs](https://rollupjs.org/guide/en/#outputexports) could be a factor if `default` option is inferred:
> default – if you are only exporting one thing using export default ...; note that this can cause issues when generating CommonJS output that is meant to be interchangeable with ESM output, see below

I've tried various alterations to the `Microbundle` setup and `tsconfig` and only managed to get predictable results by switching `src/index.js` from a `default export` to a `named export`: e.g.
```cjs
const { OpenProps } = require('open-props')
```
```mjs
import { OpenProps } from 'open-props'
```
I validated this update by:
- `npm link`ing `open-props` to a local `test-open-props-types` project
- Creating a `.mjs` and `.cjs` file that named import `OpenProps` and log `OpenProps.animationBlink`
- Ran both scripts verifying the logs
- Ran a type check with the minimum `compilerOptions` `npx tsc <file> --noEmit --allowJs --checkJs`

### IMPORTANT: This would be a breaking change! 

Hopefully, someone more familiar with `microbundle` and hybrid `ESM/CommonJs` packages can chime to avoid introducing a breaking change. However, as mentioned above converting to `named exports` was the only success I've had.

argyleink (commented):
nice, leaning on microbundle is a good strategy here 👍🏻 

during testing I couldnt figure out why/how the typed export is missing all the longform string accessors like `op["--size-1"]` as only `op.size1` works. looking at the generated `index.d.ts` I can see that the string versions are missing. the JS api should support both object and array accessor styles, and I wasnt able to find a swift resolution in 15m. thoughts on this?

Regarding `default` on the CJS export, this has been an area of frustration, and I've done lots of iterating to get it to the current state. I'd prefer to not create any breaking changes, let's continue iterating here to see what we can do.

there's [another engineer working to add types in a different fashion](https://github.com/argyleink/open-props/pull/223), maybe y'all can work together and find the best way to provide types without requiring changes to the current output?

src/index.js
```diff
@@ -18,13 +18,18 @@ const camelize = text => {
   return text.substr(0, 1).toLowerCase() + text.substr(1)
 }
 
-const mapToObjectNotation = props => {
+/**
+ * @template T
+ * @param {T} props
+ * @returns {import("type-fest").CamelCasedPropertiesDeep<T>}
```
hchiam, line 24:
```suggestion
 * @returns {import("type-fest").CamelCasedPropertiesDeep<T> & import("type-fest").KebabCase<T>}
```
to get both `op.size1` and `op["--size-1"]` to work, to address part of @argyleink's feedback https://github.com/argyleink/open-props/pull/225#pullrequestreview-944673584 

I'll keep investigating [my alternative solution](https://github.com/argyleink/open-props/pull/223), but I'm also liking the approach in this PR 😃

hchiam (changes requested):

aaronccasanova:
> the JS api should support both object and array accessor styles, and I wasnt able to find a swift resolution in 15m. thoughts on this?

It's a quick update https://github.com/argyleink/open-props/pull/225/commits/bd34e2f39c9add0cf1afc04797cd20c5901b281f. For some reason I thought the original utility was replacing the dashed properties with camel case keys. The fix is simply intersecting the original inferred type `T`.

image


> I'd prefer to not create any breaking changes, let's continue iterating here to see what we can do.

That's what I figured, makes sense 👍 Might just have to introduce the `typescript` dev dependency and run `tsc` with the `--emitDeclarationsOnly` option. I'll play around with `microbundle` a bit more and otherwise fallback to `tsc`.

src/index.js
aaronccasanova: Thanks for the suggestion! `T` already infers the kebab keys from the original type, so I just needed to intersect `T` with the camel cased type: https://github.com/argyleink/open-props/commit/bd34e2f39c9add0cf1afc04797cd20c5901b281f

hchiam:
Here's a repo I used to test autocomplete and typings for this PR: 

https://github.com/hchiam-test-org/svelte-kit-typescript-open-props2/blob/main/my-app/package.json#L37 (WIP) 

cf. https://github.com/argyleink/open-props/pull/223#issuecomment-1126590623

aaronccasanova:
@argyleink I haven't had the bandwidth to debug the ESM/CJS interop issues! Feel free to close the PR and/or @hchiam you're welcome to take it over!

Jarvis1010:
does this still need help?  I would be happy to pick it up since I would love to get the Types working as first-class citizens

argyleink:
available in `v1.5.4`!

PR #226, opened 2022-04-19
Add unnecessary config files to the `.npmignore` & change logo alt text

nucliweb:
## In this pull Request

> Based on [comment](https://github.com/argyleink/open-props/pull/224#issuecomment-1101564351) 

- [x] Added `firebase.json`and `postcss.config.js` to the `.npmignore` file
- [x] Change the `alt` text to the Open Props logo in the readme file

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,6 @@
 .firebaserc
 .github
+firebase.json
+postcss.config.js
 docsite/
 test/
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
 .firebaserc
 .github
 firebase.json
-postcss.config.cjs
 docsite/
 test/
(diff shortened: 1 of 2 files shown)


.npmignore
```diff
@@ -1,4 +1,6 @@
 .firebaserc
 .github
+firebase.json
+postcss.config.cjs
```
argyleink, line 4: makes sense to omit firebase config, but i think leaving the postcss config makes sense in case someone npm installs this and wants to build it themselves?

nucliweb: IMO, in this case, you can use the source code, right?
Is a strange use case to build a package in side the `node_modules` folder

Any way, I was remove the `postcss.config.cjs`

argyleink (approved):

Issue #212, opened 2022-03-23
<form> focus style is clipped and jagged in Safari


argyleink:
https://user-images.githubusercontent.com/1134620/159728510-4f7fd6ea-2ecf-443c-81c8-840f551ccdab.mp4

mrtuvn:
outline: 0 for input that not disabled may help for such case

argyleink:
`wont fix`, see https://github.com/argyleink/open-props/pull/244#issuecomment-1179653685

argyleink:
fixed in Safari Tech Preview

PR #244, opened 2022-07-03
Fix outline focus at Safari

mrtuvn:
https://github.com/argyleink/open-props/issues/212

### DIFF
--- a/src/extra/normalize.css
+++ b/src/extra/normalize.css
@@ -40,7 +40,12 @@
 :where(a, button, input, textarea, summary) {
   --outline-size: max(5px, 1px);
   --outline-style: solid;
-  --outline-color: var(--blue-5);
+  --outline-color: currentColor;
+}
+
+:where(a, button, input, textarea, summary):focus {
+  outline: var(--outline-size) var(--outline-style) var(--outline-color);
+  outline-offset: var(--outline-offset, var(--outline-size));
 }
 
 :where(a, button, input, textarea, summary):focus-visible {
--- a/src/extra/normalize.css
+++ b/src/extra/normalize.css
@@ -37,7 +37,11 @@
   outline-offset: 5px;
 }
 
-:where(body) { 
+:where(input[type='text']:not([disabled]):focus) {
+  outline: none;
+}
+
+:where(body) {
   min-block-size: 100%;
 }
 
@@ -46,13 +50,13 @@
   font-weight: var(--font-weight-9);
 }
 
-:where(h1) { 
-  font-size: var(--font-size-8); 
-  max-inline-size: var(--size-header-1); 
+:where(h1) {
+  font-size: var(--font-size-8);
+  max-inline-size: var(--size-header-1);
 }
 
-:where(h2) { 
-  font-size: var(--font-size-6); 
+:where(h2) {
+  font-size: var(--font-size-6);
   max-inline-size: var(--size-header-2);
 }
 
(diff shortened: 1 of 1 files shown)


src/extra/normalize.css
```diff
@@ -37,7 +37,11 @@
   outline-offset: 5px;
 }
 
-:where(body) { 
+:where(input[type='text']:not([disabled]):focus) {
+  outline: none;
+}
```
argyleink, line 42: sorry, the issue is bigger than just text inputs, it's just what is shown in the video. were you able to reproduce the issue on the text input?

src/extra/normalize.css
```diff
@@ -46,13 +50,13 @@
   font-weight: var(--font-weight-9);
 }
 
-:where(h1) { 
-  font-size: var(--font-size-8); 
-  max-inline-size: var(--size-header-1); 
+:where(h1) {
+  font-size: var(--font-size-8);
+  max-inline-size: var(--size-header-1);
```
argyleink, line 55: can you reduce the PR to only commits needed to fix the issue/feature please? 🙂

argyleink (changes requested):
solution needs to be larger than just text inputs

src/extra/normalize.css
```diff
@@ -37,6 +37,21 @@
...
 
+:where(a, button, input, textarea, summary) {
+  --outline-size: max(5px, 1px);
+  --outline-style: solid;
+  --outline-color: var(--blue-5);
+}
+
+:where(a, button, input, textarea, summary):focus-visible {
+  outline: var(--outline-size) var(--outline-style) var(--outline-color);
+  outline-offset: var(--outline-offset, var(--outline-size));
```
argyleink, line 48: where is `--outline-offset` defined?

src/extra/normalize.css
```diff
@@ -37,6 +37,21 @@
   outline-offset: 5px;
 }
 
+:where(a, button, input, textarea, summary) {
+  --outline-size: max(5px, 1px);
```
argyleink, line 41: can you explain this a bit? how'd you come across it?

argyleink (commented):
nice, ty for the revisions! 
i've added a couple questions and have done some light testing. 
i'm curious how you're testing the changes and verifying?

src/extra/normalize.css
mrtuvn: added some markup at docsite directory for testing from my mac

src/extra/normalize.css
mrtuvn: if --outline-offset missed or not define by user the second param will used instead

src/extra/normalize.css
mrtuvn: just add some local custom variables for easier to understand

src/extra/normalize.css
mrtuvn: In case redefine offset we can do like this
```css
input:focus-visible {
  --outilne-offset: 5px;
}
```

argyleink:
`max(5px, 1px)` works because it sets the base value to the same as the focus value `5px`, really not transitioning any distance anymore and therefore not reproducing the bug. 

i think this is a `wont fix` in open props, instead i'll log the bug on webkit's bug repo to see if they'll fix it. like, if you hover over the poorly painted animated outlined element, the outline fixes itself. seems like a paint / animation bug that we can get fixed upstream for everyone instead of patching it with a custom style.

i appreciate you hackin on this and making the PR! ❤️  it forced me to investigate the issue further and establish a stronger understanding of the problem.

PR #258, opened 2022-08-09
Optimization for the mask edges

Afif13:
Found better implementation for the edge masks. I feel a bit stupid that I didn't find them before 😣 many of them can be reduced to only one gradient!

It seems that looking at my own code after too long is beneficial.


argyleink (approved):

argyleink:
awesome, thanks for contributing the optimizations!

PR #262, opened 2022-08-10
Optimizing the Scalloped edges

Afif13:
Yes I am addicted to shorten codes ...

I think it's gonna be the last optimization so you can update the file size on the doc.


argyleink (approved):
rad

PR #268, opened 2022-08-16
color-scheme adaptive fade-in/out-bloom

hchiam:
1st draft follow-up to https://github.com/argyleink/open-props/pull/260#issuecomment-1212490240

### DIFF
--- a/build/to-stylesheet.js
+++ b/build/to-stylesheet.js
@@ -5,8 +5,11 @@ export const buildPropsStylesheet = ({filename,props}, {selector,prefix}) => {
 
   let appendedMeta = ''
 
-  if (filename.includes('shadows')) {
+  if (filename.includes('shadows') || filename.includes('animations')) {
     file.write(`@import 'props.media.css';\n\n`)
+  }
+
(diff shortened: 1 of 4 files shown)


hchiam (commented):
Ready for review. Hopefully close to what you had in mind, @argyleink: 
color-scheme adaptive demo 

Quick reference/reminder of how all these "bloom fades" started: 
https://twitter.com/shuding_/status/1552438750470340610?s=20&t=95DjzncWa9CBSwEJkLBCbQ 
I liked how there was contrast between the "flash of light" in dark mode, but it might look out of place if we do the opposite (i.e. "darkness") in light mode. So instead of being asymmetrical/opinionated by treating dark mode differently from light mode ("flash of light" in dark mode but smoothly fade in light mode), I went with treating both similarly (smoothly fade in dark mode and smoothly fade in light mode). Thoughts?

argyleink (approved):
very cool, that's the type of result i was expecting / hoping for too. nice work!

thanks for hacking in the dark props handling in the build file 👍🏻

PR #272, opened 2022-09-01
🚀 Astro docs site

davideast:
# Astro site for OP!

## Philosophy
All components are in `.astro` format and use CSS Modules for encapsulation.

### Pros: 
- Encapsulation 
- Composes well with global utility classes
- Portable across any framework

### Cons:
- Not single file (is that a con?)
- Non-spec CSS Module syntax

What are your thoughts @argyleink?


davideast:
Coming back up for air. Stay tuned.

argyleink:
closing due to so much time passing

PR #279, opened 2022-09-16
fixes #277

### LINKED ISSUE #277: @custom-media intersections
### Discussed in https://github.com/argyleink/open-props/discussions/275

<div type='discussions-op-text'>

<sup>Originally posted by **WebMechanic** September 13, 2022</sup>
Hello,

we use `@custom-media` with the PostCSS plugin for or .scss files (Vite), and I was wondering about the reasoning behind intersecting values of the various `*-only`, `*-n-below` and `*-n-above` due to the use of `<=` and `>=`

```css
@custom-media --lg-n-above (width >= 1024px);
@custom-media --lg-n-below (width <= 1024px);
```
transpiles into
```css
@media (min-width: 1024px) ...
@media (max-width: 1024px) ...
```
Rulesets would then overlap at 1024px letting the cascade win which isn't always possible to control (not using `@layer`).

Maybe I'm missing the logic here, but wouldn't this be more accurate?
```css
@custom-media --lg-n-above    (width >= 1024px);
@custom-media --lg-n-below    (width <  1024px);  // == (max-width: 1023px)
```

So, just curious.

---

In our copy we eventuall replaced all `<=` with `<` and also used more accurate/modern screensizes based on our analytics and matching our content's needs, not the (outdated) "iDevice" values even Apple no longer maintains.
</div>

argyleink:
cc @WebMechanic 👍🏻

### DIFF
--- a/docsite/index.html
+++ b/docsite/index.html
@@ -2971,12 +2971,12 @@
 
               @custom-media --md-only       (480px &#60;= width &#60;= 768px);
               @custom-media --md-n-above    (width &#62;= 768px);
-              @custom-media --md-n-below    (width &#60;= 768px);
+              @custom-media --md-n-below    (width &#60; 768px);
               @custom-media --md-phone      (--md-only) and (--portrait);
 
               @custom-media --xxl-only      (1440px &#60;= width &#60;= 1920px);
               @custom-media --xxl-n-above   (width &#62;= 1920px);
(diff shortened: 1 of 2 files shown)


WebMechanic:
"Viewport Vars Sample" also needs updating
https://open-props--pr279-277-custom-media-int-bnv3fbbe.web.app/#media-queries

PR #280, opened 2022-09-18
Fix #278 and extend with --firefoxMacONLY, --chromiumONLY, --webkitONLY

### LINKED ISSUE #278: @custom-media browser "detection" in props.supports
Hello,

the three `@supports` rule variables in `props.supports` are unfortunately not very reliable. 

We needed a js-free, solid **Safari only** detection, not accidently catching some Chromiums, and Safaris not seeing styles intended for Google Chrome -- which is what happens now.

The `--firefoxONLY` is the only one that works as advertised thanks to the `-moz-` prefix and Mozilla using their own enginge, Gecko [^1]

The `--safariONLY` also catches Chromium browsers (Linux) and "unbranded" WebKits (they're rare but used in e2e and headless browser test systems [^2]. This could make testing a bit more guesswork.)
It also catches older Google Chrome releases (no biggie thanks to its evergreenness).

The `--chromeONLY` also works for Safari. 
I guess Chromium is ok as it's a tad closer to Google Chrome than "plain" WebKit.

### Safer Alternatives

A safe `--safariONLY` would be `-apple-pay-button-style: inherit` which goes several versions back, and like `-moz-` for Mozilla, `--apple-` is an (old) Apple OS == Safari exclusive.

A safe `--chromeONLY`  would be `(-webkit-app-region: inherit)`. No false positive for Safari. 
This a Chromium window element property.

So to eventually even get `--webkitONLY` or `--chromiumONLY` effectivly combine them with a `not`.

Check with https://noscriptfingerprint.com/ where I found both alternatives.

Enjoy.

---

For those who don't know: WebKit, Blink, Chromium, Chrome, Edge, and Safari share a lot of their code through the WebKit project. Some of these names are so called browser engines, some are browser apps people use on their devices. 
Most notably: they are **never** on par with each other in terms of CSS features.
We usually want to test for the browser app people use, 'cos vendors add and remove features at will from the (shared) engines they use for all sorts of reasons.

[^1] based on Gecko does not mean it's a (recent) Firefox though. There's also Seamonkey, K-Meleon, Pale Moon, and Fennec for Android.
[^2] Playwright on Windows.

WebMechanic:
Test page and explainer https://codepen.io/WebMechanic/pen/poVexYp?editors=1100

Sources: 
- WebKit https://webkit.org/css-status/#?status=non-standard
- Lea Verou's https://css3test.com/  "Experimental" filter
- https://github.com/fingerprintjs/blog-nojs-fingerprint-demo/

`container-type:`**`none`** is a non-standard value to test against. It **must not** be changed to `inherit` which is usually sufficient and safe to check property support.

### DIFF
--- a/src/props.supports.css
+++ b/src/props.supports.css
@@ -1,6 +1,9 @@
-@custom-media --safariONLY  (-webkit-hyphens: none);
-@custom-media --firefoxONLY (-moz-appearance: none);
-@custom-media --chromeONLY  (-webkit-tap-highlight-color: black);
+@custom-media --safariONLY      (-apple-pay-button-style: inherit);
+@custom-media --firefoxONLY     (-moz-appearance: inherit);
+@custom-media --firefoxMacONLY  (-moz-osx-font-smoothing: inherit);
+@custom-media --chromeONLY      (-webkit-app-region: inherit) and (not (container-type: none));
+@custom-media --chromiumONLY    (-webkit-app-region: inherit) and (container-type: none);
+@custom-media --webkitONLY      (alt: inherit) and (not (-apple-pay-button-style: inherit))
 
 :where(html) {
   --isLTR: 1;


argyleink (approved):
so far my tests have gone good and the updates seem fair. it is odd seeing `container-type: none` in there, but the reasoning seems good. 

going to test more at work. thanks for this!

argyleink:
where did you learn that `container-type:none` matched? 
i'm on the fence about the chromium one, you?

WebMechanic:
> where did you learn that container-type:none matched?

using Lea Verou's https://css3test.com/ ("Experimental" filter selected) in **Containment Level 3 / Properties**.
It was "green" in Chromium but not in Chrome.

This is Chrome 105 (and WebKit "16")

image

> i'm on the fence about the chromium one, you?

It could be useful to tweak font stuff but that's all I can think of now :)

That's likely the test that might break[^1] sooner or later[^2]. I wouldn't mind, if you want to skip it. It was more of a challenge to find sth _at all_ and some sort of a "first match pick" from Lea's site. There could be others 🤷‍♂️ .

To be frank: we virtually have no Chromium or Linux visitors in our analytics so we don't really test in that browser. If stuff works in Chrome it's fair to presume it works in Chromium too. Both are very much on par with their CSS for "real world work".

I wouldn't add it back in for our copy. It's now "documented" and if somebody needs different style for Chrome and Chromium, here's how they learn how :D or scan css3test.com for sth else.

Your descision after all.

[^1] I just saw that Firefox' experimental implementation parses **all** properties here = it's all green.
[^2] 'none' appears to be from [the 20211221 draft](https://www.w3.org/TR/2021/WD-css-contain-3-20211221/#container-type). It has since been [removed](https://www.w3.org/TR/css-contain-3/#container-type). So should Chromium **update** their implementation it might be gone, too.

Issue #286, opened 2022-09-21
Include theme toggling in normalize.css


Crisfole:
I spent a good deal of time trying to figure out why `data-theme=light` and `data-theme=dark` didn't work in normalize.

As it turns out, it's not included at all; I think it'd be pretty reasonable to add to `normalize.css` and largely would just mean duplicating all the `@media (--OSDark)` and `@media (--OSLight)` rules with the `@nest [data-theme=dark] &` and `@nest [data-theme=light] &` rules.

Last I checked CSS does not allow alternation between `@media` rules and other selectors, but POSTCSS might have something for that? That would make it even easier.

argyleink:
good call, this should totally be provided in the package for easy use 👍🏻 

what do you think of just adding 2 more imports?

```css
@import "open-props";
@import "open-props/normalize";
@import "open-props/theme/light";
@import "open-props/theme/dark";
```

those theme imports would bring in just the `[data-theme="dark"]` duplication needed to have clientside exceptions. This was it doesn't impact any of the current files and is an additive feature for when folks want to deploy a theme toggle? I could add it to [this PR](https://github.com/argyleink/open-props/pull/285) which is adding "light only" or "dark only" normalize options. Or, you could continue with your PR, add the 2 new files and that'd be it?!

argyleink:
fixed in v1.4.17 👍🏻

Crisfole:
Is this documented? Did you stick with “theme”?

argyleink:
it's not really documented yet nope. i should make a demo and then add some prose to the docsite. will do soon in #304

`@import "open-props/switch/light";` or `@import "theme.light.switch.min.css";`
`@import "open-props/switch/dark";` or `@import "theme.dark.switch.min.css";`

PR #287, opened 2022-09-21
Stab at adding data-theme-dark respect

Crisfole:
This is _very_ much a stab-in-the-dark, first-pass effort. Please don't merge, please don't judge. I'm 100% confident I'm missing `&` in a few spots, but I'm also pretty sure it's not far off.

I also have to start work, but I didn't want to drop an Issue (#286) without a proposal. Feel free to close both this and #286 without fearing I'll whine at you.

### DIFF
--- a/src/extra/buttons.css
+++ b/src/extra/buttons.css
@@ -6,42 +6,42 @@
   --_accent: initial /* your color */;
   --_text: initial /* your text color */;
   --_size: initial /* your size */;
-  
+
   --_bg-light: white;
   --_bg-dark: var(--surface-3);
   --_bg: var(--_bg-light);
-  
+
   --_border: var(--surface-3);
-  
+
   --_highlight-size: 0;
   --_highlight-light: hsl(210 10% 71% / 25%);
   --_highlight-dark: hsl(210 10% 5% / 25%);
   --_highlight: var(--_highlight-light);
-  
+
   --_ink-shadow-light: 0 1px 0 var(--gray-3);
   --_ink-shadow-dark: 0 1px 0 var(--surface-1);
   --_ink-shadow: var(--_ink-shadow-light);
-  
+
   --_icon-size: 2ch;
   --_icon-color: var(--_accent, var(--link));
-  
+
   font-size: var(--_size);
   background: var(--_bg);
   color: var(--_text);
   border: var(--border-size-2) solid var(--_border);
-  box-shadow: 
+  box-shadow:
     var(--shadow-2),
     0 1px var(--surface-3),
     0 0 0 var(--_highlight-size) var(--_highlight)
   ;
   text-shadow: var(--_ink-shadow);
-  
+
   display: inline-flex;
   justify-content: center;
   align-items: center;
   text-align: center;
   gap: var(--size-2);
-  
+
   font-weight: var(--font-weight-7);
   border-radius: var(--radius-2);
   padding-block: .75ch;
@@ -50,7 +50,7 @@
   user-select: none;
   -webkit-tap-highlight-color: transparent;
   -webkit-touch-callout: none;
-  
+
   transition: border-color .5s var(--ease-3) 3s;
 
   @media (--OSdark) {
@@ -59,8 +59,14 @@
     --_ink-shadow: var(--_ink-shadow-dark);
   }
 
+  [data-theme=dark] {
+    --_highlight: var(--_highlight-dark);
+    --_bg: var(--_bg-dark);
+    --_ink-shadow: var(--_ink-shadow-dark);
+  }
+
   @media (--motionOK) {
-    transition: 
+    transition:
       border-color .5s var(--ease-3) 3s,
       box-shadow 145ms var(--ease-4);
   }
@@ -75,6 +81,10 @@
     box-shadow: var(--shadow-1);
 
     @media (--OSdark) {
+      --_text: var(--gray-5);
+    }
+
+    [data-theme=dark] {
       --_text: var(--gray-5);
     }
   }
@@ -103,7 +113,7 @@
 /* adaptive indigo text */
 :where([type="submit"], form button:not([type],[disabled])) {
   --_text: var(--_accent, var(--link));
-} 
+}
 
 /* red reset */
 :where([type="reset"]) {
@@ -115,6 +125,11 @@
   }
 
   @media (--OSdark) {
+    --_text: var(--red-2);
+    --_border: var(--surface-3);
+  }
+
+  [data-theme=dark] {
     --_text: var(--red-2);
     --_border: var(--surface-3);
   }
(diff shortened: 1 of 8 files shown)


argyleink:
thanks for making this, it helps a lot with our conversation!

src/extra/buttons.css
```diff
@@ -6,42 +6,42 @@
   --_accent: initial /* your color */;
   --_text: initial /* your text color */;
   --_size: initial /* your size */;
-
```
argyleink, line 9: some linter is making this PR look like a much bigger change than it needs to be! can you commit only the lines that need to change please 🙂

src/extra/buttons.css
```diff
@@ -59,8 +59,14 @@
     --_ink-shadow: var(--_ink-shadow-dark);
   }
 
+  [data-theme=dark] {
```
argyleink, line 62: nice attention to detail in the repo here, noticing buttons including dark theme related selectors 👍🏻 

```suggestion
  @nest [data-theme=dark] & {
```

that's the selector we need here, so that any ancestor with that attribute and value pair then style buttons 👍🏻 

how comes `--_bg` and `--_highlight` appear to be added?

src/extra/buttons.css
```diff
@@ -147,7 +162,16 @@
...
   :where(form button:not([type="button"])),
-  :where([type="reset"]), 
+  :where([type="reset"]),
+  :where([disabled]) {
+    --_bg: var(--surface-1);
+  }
+}
+
+[data-theme=dark] {
+  :where([type="submit"]),
```
argyleink, line 172:
```suggestion
  & :where([type="submit"]),
```

syntax requires an `&` to signal the nesting 👍🏻  each of these here need an & in front of them, or chained all within the same `:where()`

src/extra/normalize.css
```diff
@@ -135,6 +135,10 @@
   @media (--OSdark) {
     background-color: hsl(210deg 11% 10%);
   }
+
+  @nest [data-theme=dark] {
```
argyleink, line 139:
```suggestion
  @nest [data-theme="dark"] & {
```

src/extra/normalize.css
```diff
@@ -283,6 +287,10 @@
     background-color: var(--surface-2);
   }
 
+  @nest [data-theme=dark] & {
```
argyleink, line 290: nice! looks like you found your way. 

for consistency, can you put the attribute values in quotes?

```suggestion
  @nest [data-theme="dark"] & {
```

src/extra/theme.dark.css
```diff
@@ -13,4 +13,19 @@
...
+    --link: var(--indigo-3);
+    --link-visited: var(--grape-3);
+
+    --text-1: var(--gray-1);
+    --text-2: var(--gray-4);
+
+    --surface-1: var(--gray-9);
+    --surface-2: var(--gray-8);
+    --surface-3: var(--gray-7);
+    --surface-4: var(--gray-6);
```
argyleink, line 30: i'm thinking of making this into a file, letting someone import it like `@import "open-props/theme-switch/dark";`. These other spots with dark preference media queries also need duplicated though too, i wonder if they should be in the same file? 🤔

src/extra/theme.light.css
```diff
@@ -19,3 +19,25 @@
     }
   }
 }
+
+[data-theme=light] {
```
argyleink, line 23: i feel like the data attribute styles should be opt-in, folks knowing their making a switch

src/props.animations.css
```diff
@@ -120,10 +120,24 @@
   100% { opacity: 1; filter: brightness(1) blur(0) }
   };
 }
+[data-theme=dark] {
```
argyleink, line 123: nice find 👍🏻

src/props.shadows.css
```diff
@@ -47,4 +47,11 @@
     --shadow-color: 220 40% 2%;
     --shadow-strength: 25%;
   }
+}
+
+[data-theme=dark] {
```
argyleink, line 52: this needs to happen here, a place i need to refactor, but here nonetheless: https://github.com/argyleink/open-props/blob/main/build/to-stylesheet.js#L28-L32

add a new line with the `[data-theme="dark"]` selector and put `${dark_propsMeta}` inside.

argyleink (changes requested):
we can work with this if you want to carry it across the finish line?
it's done much of research and work already and would make a great reference for me to finish in the other normalize branch i'm working on. if you want, i can pick it up?
lemme know!

Crisfole:
Comments are all great! I’ll take a look this morning. Separate file makes a ton of sense. I kept them together to keep them in sync…if you change one it forces changing another.

src/extra/buttons.css
Crisfole: Huh. I didn't see those changes...Weird. I'll reset HEAD^ and git add -p...

src/extra/buttons.css
Crisfole: I have to assume my editor was formatting on save, and there's no formatter setup that VSCode could find...

src/extra/buttons.css
Crisfole: I figured out why my editor made all these changes! You have trailing spaces all over ;). Any interest in a code formatter to get rid of all these silly nits?

image

src/extra/normalize.css
Crisfole: 🤦 I even got halfway there! 😆 First time using the proposed nest syntax. I don't hate it, but I do hate the inconsistency between it and the `@media` syntax.

src/extra/theme.dark.css
Crisfole: This is definitely a great plan. I'm mildly concerned with the dark preference duplication being hard to manage across multiple files which is why I started with including it in all situations.

If _every_ use of `@media (--OSDark)` or `@media (--OSLight)` was in the theme file I'd agree with you 100%.

Crisfole: You could *pretty* easily write an automatic refactor-er (real word, promise!) that would split the media queries for OSDark and OSLight into two and send them into two different files.

src/extra/theme.light.css
Crisfole: Yep. Shipping extra code should be opt-in. That said, I think you should probably _also_ include a 'all the bells and whistles' option you can import to get everything.

src/props.animations.css
Crisfole: Instead of using selectors and media this shoudl be using variables so that you can keep the theme in one place.

Crisfole:
Closing in favor of #289

src/extra/theme.dark.css
argyleink: def agree there's ways to automate this AND agree it's not fun to manage the media queries + attributes. there's a css proposal that would help here a lot called [`:media()`](https://github.com/w3c/csswg-drafts/issues/6247) that would allow combining the attribute selector and the media query into 1 selector, but it's only a proposal.

argyleink:
sounds good, will check back in when there's new files to import from the published package 👍🏻

PR #289, opened 2022-09-22
Feature/compact data theme dark

Crisfole:
This re-opens #287 without all the cruft and _with_ all the feedback except splitting theme into its own file.

I assume this will be a helpful reference for @argyleink


argyleink:
very helpful indeed

argyleink:
fixed in v1.4.17 👍🏻 
give it a shot!

PR #290, opened 2022-09-22
Feature/formatting

Crisfole:
Do not read this PR as a whole. Read it a commit at a time. Consider *not* reading 'Run the formatter'.

This is a big change to most of the files, so also consider merging when you have closed or gotten close to closing most of your other PRs. If you do so, you'll want to do the following:

```
git cherry-pick 08b3d26
npm run pretty
git add . && git commit -m 'Run the formatter'
```

This PR adds a code formatter (prettier) and a pre-commit hook that runs on commit on *only* the changed files. It uses the core prettier config and gets run on every file.

You might hate this, and that's fine. Please feel free to close, it didn't take me long. But it would have made random changes like the trailing whitespaces non-issues. It would make PRs easier to review


argyleink:
thanks for making a PR that intends to aid in commits by enforcing a style, but these generally aren't my cup of tea. maybe i should add an editorconfig, that's what i've done in the past, that'll help your default linters stop changing everything in this project. i'd also have gone with [standard](https://standardjs.com/) if i was to pick one, but yeah, not ready to enforce anything like this yet. thanks tho!

Crisfole:
Absolutely no problem. Editor config would definitely help. At this point I'm just keeping a .vscode workspace in my repo with the right overrides.

PR #293, opened 2022-10-08
Update VScode instructions

diegoramosxyz:
- Fixed typo from 'when' to 'where'
- Removed unsupported file extension `html` from CSS Var Complete configuration


argyleink (approved):
🙏🏻

PR #297, opened 2022-10-20
Fix docs variable names (surface & text)

jorgecasar:


argyleink:
oh, i see you're matching the normalize variable names with the ones in the color theming section?

docsite/index.html
```diff
@@ -1497,149 +1497,9 @@ <h5>Prefers Dark</h5>
...
-              </code></pre>
-            </div>
-            <div>
-              <h5>Prefers Dim</h5>
-              <pre class="language-css"><code>
-                @media (prefers-contrast: less) {
-                  :root {
-                    --brand: var(--brand-dim);
-
-                    --text1: var(--text1-dim);
```
argyleink, line 1518: how come this whole chunk is deleted?

jorgecasar: Sorry, I edited online an maybe some copy and paste was the cause. I will fix it

jorgecasar:
Yes, I see that `--text1` doesn't exits, it's `--text-1` instead. And same with surfaces.

argyleink:
i'm torn here. but i don't think we need to align the color guide syntax with normalize. i appreciate the thought about consistency, but it's just a guide and not necessarily syntax shipping in the lib.

i'm going to close for now, but if it comes up again, i'll reopen and we can align everything. thanks tho!!

PR #301, opened 2022-10-28
animate outline offset for buttons

fauxparse:
I noticed in your [intro video](https://www.youtube.com/watch?v=9VXR_qRgROE) (thanks, btw!) that you were puzzled as to why the `outline-offset` on your buttons wasn't animating. I think this fixes it?


argyleink (approved):
yep 🤘🏻 💀

PR #309, opened 2022-11-05
Fix range padding

fauxparse:
The default `padding-inline` on a `range` input meant that the thumb was unable to be positioned at the ends of the range track.


src/extra/normalize.src.css
```diff
@@ -118,7 +118,7 @@
   letter-spacing: inherit;
 }
 
-:where(input, textarea) {
+:where(input:not([type="range"]), textarea) {
```
argyleink, line 121: yep, perfect 👍🏻

argyleink (changes requested):
this looks great! 
but a linter has made the PR include a bunch of space adjustment junk. can you get the PR to just the 1 line change please? then this is good to go 🎉

argyleink:
available in `v1.4.24` releasing now 👍🏻

PR #314, opened 2022-11-16
Migrate from `open-color` to `colar`

kelvindecosta:
## Overview

This PR attempts to migrate from using `open-color` to `colar`.
The latter is based on the former and extends the color palette.

Screenshot_20221116_220142

## Notes

Here are some changes that I thought were worth highlighting:

- Replace `--grape` with `--purple` (palette values are unchanged)
- Add `colorjs.io`, required by `props.colors.src.js`, as a `devDependency` as its CDN seems to be unavailable
- Set `.color-swatch` `inline-size` as ~~`7%`~~ `6.7%`, to account for `13` luminosity steps.

I decided not to make changes to any theming (including the documentation).
It could use some love:

Screenshot_20221116_220402

Screenshot_20221116_220428

## Related Issues

Should fully/partially address #247 & #308

## Conclusion

I tried to be as systematic and careful as possible.

Any feedback is appreciated!

PR #317, opened 2022-11-20
v1.5.0

### LINKED ISSUE #247: More dark colors
It's the only thing about Open Color I don't like so far is the lack of a few darker colors. Well, turns out I'm not alone and the work has been done here! https://github.com/fchristant/colar

Need to decide if these should ship sooner than later, or wait for relative color syntax to make this a dynamic task instead of one with all the presets. 🤔 

https://github.com/yeun/open-color/issues/119

argyleink:
* Migrate from `open-color` to `colar` (#314)
* feat: update usage of `--grape` to `--purple`
* feat: new `stone` palette
* fixes #247


argyleink:
there's still color work to be done, but this is a step forward

PR #318, opened 2022-11-20
 Package subpath './open-props.min.css' is not defined by "exports"

soluml:
Looks like the main docsite needs an update here as the latest package exports `open-props.min.css` as "style".

### DIFF
--- a/docsite/index.html
+++ b/docsite/index.html
@@ -1050,7 +1050,7 @@
                       plugins: [
                         postcssJitProps({
                           files: [
-                            require('open-props/open-props.min.css'),
+                            require('open-props/style'),
                           ]
                         }),
                       ]


soluml:
Actually, attempting to use this in an Astro site, I needed to pass the path as a string:

`files: ["node_modules/open-props/open-props.min.css"],`

argyleink:
the syntax currently on the docsite is latest NPM exports syntax, it's a bummer astro couldnt understand it in your case. i'm going to keep it at the current syntax. module resolution is confusing stuff tho, sorry you had to debug this.

soluml:
Oh, yea, sorry I wasn't clear. I ran into two problems. The first one was an exports issue, the stylesheet is set as "style" in package.json. 

https://github.com/argyleink/open-props/blob/main/package.json#L42

The second issue may be related to Astro and its mjs config leveraging posgcss as a cjs config.

argyleink:
the `style` in package.json is used by PostCSS when `@import "open-props";` is used bare, it knows to resolve the style module due to the style context of the import. it shouldnt be blocking access to any files, it's just an alias. `./open-props.min.css` is exported via the `/style` export but is also always available directly without any alias (as you've found). 

yeah, the hell of modules are require and imports and alias resolution etc.. it's confusing work.

PR #323, opened 2022-11-28
update autocomplete settings

equinusocio:
Currently that extension's setting works on language ID's and not file extensions. If the consumer project is using postcss, it must be specified in the setting to enable autocompletition on this language/files

Ref. https://github.com/willofindie/vscode-cssvar/issues/99


argyleink (approved):
nice, thanks for helping others out by updating these docs 🙂

equinusocio:
I just spent an hour figuring out why it was not working on my postcss files...

PR #326, opened 2022-12-16
Add missing exports for masks and media js files

mayank99:
added `media.js`, `masks.edges.js` and `masks.corner-cuts.js` to exports because they were previously missing and impossible to import.

unsure if this is the right convention (i couldn't find any previous examples with multiple dots in the file name)

also, i didn't include it in this pr but it might help to add a catch-all to the export map. this will make sure everything is exposed (unless some things are purposely hidden).
```js
"./": "./"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,8 @@
     "./src/sizes": "./src/props.sizes.js",
     "./src/svg": "./src/props.svg.js",
     "./src/zindex": "./src/props.zindex.js",
+    "./src/masks.edges": "./src/props.masks.edges.js",
+    "./src/masks.corner-cuts": "./src/props.masks.corner-cuts.js",
     "./style": "./open-props.min.css",
     "./postcss/style": "./src/index.css",
     "./normalize": "./normalize.min.css",
@@ -167,7 +169,8 @@
     "./postcss/theme": "./src/extra/theme.css",
     "./postcss/theme-dark": "./src/extra/theme-dark.css",
     "./postcss/theme-light": "./src/extra/theme-light.css",
-    "./utilities": "./src/extra/utilities.css"
+    "./utilities": "./src/extra/utilities.css",
+    "./*": "./*"
   },
   "browserslist": [
     "defaults"


package.json
```diff
@@ -39,6 +39,9 @@
     "./src/sizes": "./src/props.sizes.js",
     "./src/svg": "./src/props.svg.js",
     "./src/zindex": "./src/props.zindex.js",
+    "./src/media": "./src/props.media.js",
```
argyleink, line 42: this was already exposed as `"./esm/media": "./src/props.media.js",` 👍🏻

argyleink (changes requested):
thanks for catching and adding those 🙂 
this looks good to me except the duplicate media entry. 

Side thought:
maybe i should make a separate issue to normalize the naming convention so all the js modules use the esm/ prefix 🤔 

Side Q:
what env are you in that you can't reach into the package and grab any file you want? which env is forcing you only to importing exported items? `import "open-props/docsite/syntax-highlight.css";` should just work if someone wants it to?

package.json
mayank99: ooh, i missed that. undone! 👍

if you would like everything to be under `esm/`, i can update the other two as well.

the documentation mentions `src/` fwiw

mayank99:
> Side thought: maybe i should make a separate issue to normalize the naming convention so all the js modules use the esm/ prefix 🤔

makes sense to me! but the old export paths would need to be kept around to avoid breaking changes.

> Side Q: what env are you in that you can't reach into the package and grab any file you want? which env is forcing you only to importing exported items? `import "open-props/docsite/syntax-highlight.css";` should just work if someone wants it to?

it's just a plain node esm environment (i'm building a wrapper over open-props using an [`.mjs` script](https://github.com/mayank99/open-props-scss/blob/main/package/generateScss.mjs)).
```js
import MaskEdges from 'open-props/src/props.masks.edges.js';
```
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]:
  Package subpath './src/props.masks.edges.js' is not defined by "exports" in open-props/package.json
```

package.json
argyleink: i'll fix in #327, thanks!

argyleink:
which bundler is doing that npm lookup?

also, adding `"./": "./"` sounds good to me. i thought folks generally could already snag what they needed, but sounds like your bundler is restricted to whats in the exports, or limited, either way it's a bummer for you when the files are right there on disk..

argyleink (approved):
can include `"./": "./"` in this pr or another, no matter to me, but this one is g2g 👍🏻

package.json
mayank99: > i'll fix in #327, thanks!

that is technically a breaking change, jsyk (after you mentioned `esm/media`, i added it in my script. now if i upgrade to 1.5.2, it will break my script)

mayank99:
> which bundler is doing that npm lookup?

there is no bundler. it is just a plain node script running in ESM mode. the command i run is `node generateScript.mjs`.

if there *was* a bundler though, i do know vite/astro use plain node ESM for SSR

> can include `"./": "./"` in this pr or another, no matter to me, but this one is g2g 👍🏻

done!

package.json
argyleink: yeah, it is technically one! but i also silently released it, never documented it and have discord if anyone else finds the feature and change.

argyleink:
released in `v1.5.3` 👍🏻

PR #331, opened 2023-01-07
Fix redundant selectors in normalize

jacobrask:
Preparing a fork/variant of the "extra" styles and found some redundant selectors.

- Redundant element selectors in `input:not()` (`input` cannot also be `button`)
- Redundant `[type]` in `:where(button,button[type])` - has no effect since both selectors have 0 specificity
- Redundant nesting in `svg:not([width])`


argyleink (approved):
rad, thanks for the PR and the attention to detail 👍🏻

PR #335, opened 2023-01-20
add TS support, cleans up modules

### LINKED ISSUE #333: type definitions?

argyleink:
fixes #333 

microbundle wasnt building friendly files, making them hard to infer types from and impossible for authors to read. this drastically simplifies the build output for js modules while also adding a `.ts` file and `.d.ts` typings file.

preview of `open-props.ts`:

```ts
export default {
  "--animation-fade-in": "fade-in .5s var(--ease-3)",
  "--animation-fade-in-@": "\n@keyframes fade-in {\n  to { opacity: 1 }\n}",
  "--animation-fade-in-bloom": "fade-in-bloom 2s var(--ease-3)",
  ...
}
```

preview of `open-props.d.ts`:

```ts
declare const OpenProps: {
  "--animation-fade-in": "string";
  "--animation-fade-in-@": "string";
  "--animation-fade-in-bloom": "string";
  "--animation-fade-in-bloom-@": "string";
  ...
}
```

thoughts @mayank99 or @hchiam?

no new deps, no invasive or fancy ts; just an easier to infer dist file for TS to grab

### DIFF
--- a/build/props.js
+++ b/build/props.js
@@ -106,7 +106,7 @@ const TSD = fs.createWriteStream('../dist/open-props.d.ts')
 TSD.end(`declare const OpenProps: ${JSON.stringify(toTypes(), null, 2).replaceAll(',',';')};\nexport default OpenProps;`)
 
 preparedTypes().forEach(({filename, json}) => {
-  let bundle = fs.createWriteStream('../dist/props.'+filename+'.d.ts');
+  let bundle = fs.createWriteStream('../src/props.'+filename+'.d.ts');
   bundle.end(`declare const _default: ${JSON.stringify(json, null, 2).replaceAll(',',';')};\nexport default _default;`)
 })
 
--- a/build/props.js
+++ b/build/props.js
@@ -17,6 +17,8 @@ import MaskCornerCuts from '../src/props.masks.corner-cuts.js'
 
 import {buildPropsStylesheet} from './to-stylesheet.js'
 import {toTokens} from './to-tokens.js'
+import {toObject} from './to-object.js'
+import {toTypes} from './to-types.js'
 import {toFigmaTokens} from './to-figmatokens.js'
 
 const [,,prefix='',useWhere,customSubject='',filePrefix=''] = process.argv
@@ -81,6 +83,27 @@ FigmaTokens.end(JSON.stringify(figmatokens, null, 2))
 const figmatokensSYNC = { 'open-props': { ...figmatokens } }
 const FigmaTokensSync = fs.createWriteStream('../open-props.figma-tokens.sync.json')
 FigmaTokensSync.end(JSON.stringify(figmatokensSYNC, null, 2))
+
+if (!fs.existsSync('../dist'))
+  fs.mkdirSync('../dist')
+
+const JS = fs.createWriteStream('../dist/open-props.js')
+JS.end(`var OpenProps = ${JSON.stringify(toObject(), null, 2)}`)
+
+const ES = fs.createWriteStream('../dist/open-props.module.js')
+ES.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
+
+const CJS = fs.createWriteStream('../dist/open-props.cjs')
+CJS.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+// const UMD = fs.createWriteStream('../dist/open-props.umd.js')
+// UMD.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+const TS = fs.createWriteStream('../dist/open-props.ts')
+TS.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
+
+const TSD = fs.createWriteStream('../dist/open-props.d.ts')
+TSD.end(`declare const OpenProps: ${JSON.stringify(toTypes(), null, 2).replaceAll(',',';')}`)
 
 // gen prop variants
 Object.entries({
--- a/build/props.js
+++ b/build/props.js
@@ -18,7 +18,7 @@ import MaskCornerCuts from '../src/props.masks.corner-cuts.js'
 import {buildPropsStylesheet} from './to-stylesheet.js'
 import {toTokens} from './to-tokens.js'
 import {toObject} from './to-object.js'
-import {toTypes} from './to-types.js'
+import {toTypes, preparedTypes} from './to-types.js'
 import {toFigmaTokens} from './to-figmatokens.js'
 
 const [,,prefix='',useWhere,customSubject='',filePrefix=''] = process.argv
@@ -103,7 +103,12 @@ const TS = fs.createWriteStream('../dist/open-props.ts')
 TS.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
 
 const TSD = fs.createWriteStream('../dist/open-props.d.ts')
-TSD.end(`declare const OpenProps: ${JSON.stringify(toTypes(), null, 2).replaceAll(',',';')}`)
+TSD.end(`declare const OpenProps: ${JSON.stringify(toTypes(), null, 2).replaceAll(',',';')};\nexport default OpenProps;`)
+
+preparedTypes().forEach(({filename, json}) => {
+  let bundle = fs.createWriteStream('../dist/props.'+filename+'.d.ts');
+  bundle.end(`declare const _default: ${JSON.stringify(json, null, 2).replaceAll(',',';')};\nexport default _default;`)
+})
 
 // gen prop variants
 Object.entries({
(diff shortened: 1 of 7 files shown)


build/props.js
```diff
@@ -82,6 +84,27 @@ const figmatokensSYNC = { 'open-props': { ...figmatokens } }
...
+ES.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
+
+const CJS = fs.createWriteStream('../dist/open-props.cjs')
+CJS.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+// const UMD = fs.createWriteStream('../dist/open-props.umd.js')
+// UMD.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+const TS = fs.createWriteStream('../dist/open-props.ts')
+TS.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
```
mayank99, line 103: the TS file doesn't really provide anything. `.JS` + `.D.TS` is enough


```suggestion
```

mayank99 (commented):
so now there will be typings for the default export from root, but not for any additional paths. is the goal to stop here or to allow a separate PR to generate those additional DTS files?

hchiam (approved):
i've got nothing to add beyond @mayank99's comment https://github.com/argyleink/open-props/pull/335#discussion_r1083016282 
LGTM! :+1:

argyleink:
> but not for any additional paths. is the goal to stop here or to allow a separate PR to generate those additional DTS files?

i was going to stop here, but it's not much more effort to produce individual `.d.ts` files now the system is laid out. it'd be good to have parity in the import syntax, just so the docs can stay clean and dont diverge 👍🏻  aka, assume i'll add to this PR

build/props.js
argyleink: thanks, i was wondering about this. there's a bit more value in the .ts file than the .d.ts file because it holds the values and the value shows in the autocomplete, as opposed to just offering key completion and presenting the value type. i'll test a bit more and see what the most minimal solution is with great DX 👍🏻

argyleink:
Screenshot 2023-01-23 at 3 42 43 PM

argyleink:
alright y'all, i'll merge this tomorrow if we're feelin good about it. thanks for everyones help and effort!

argyleink:
oh! or i need to figure out why the test failed first 🙂

test/basic.test.cjs
```diff
@@ -51,4 +51,9 @@ test('Should produce normalize files', async t => {
 test('Should produce optional mask props', async t => {
   t.assert(fs.existsSync('./masks.edges.min.css'))
   t.assert(fs.existsSync('./masks.corner-cuts.min.css'))
+})
+
+test('Should produce typings files', async t => {
+  t.assert(fs.existsSync('../dist/open-props.d.ts'))
+  t.assert(fs.existsSync('../dist/props.sizes.d.ts'))
```
hchiam, line 58: i tried on stackblitz and it seems like all tests using `existsSync` are failing

build/props.js
mayank99: > there's a bit more value in the .ts file than the .d.ts file because it holds the values and the value shows in the autocomplete

but typescript cannot import anything material from `.ts` files inside node modules. it looks for `.js` and `.d.ts` files

build/props.js
```diff
@@ -82,6 +84,32 @@ const figmatokensSYNC = { 'open-props': { ...figmatokens } }
...
+
+const TSD = fs.createWriteStream('../dist/open-props.d.ts')
+TSD.end(`declare const OpenProps: ${JSON.stringify(toTypes(), null, 2).replaceAll(',',';')};\nexport default OpenProps;`)
+
+preparedTypes().forEach(({filename, json}) => {
+  let bundle = fs.createWriteStream('../dist/props.'+filename+'.d.ts');
```
mayank99, line 109: this should probably be `src` not `dist` because that's where the js files can be found
```suggestion
  let bundle = fs.createWriteStream('../src/props.'+filename+'.d.ts');
```

build/props.js
mayank99: plus, ESM environments require file extensions and/or export maps, and node would not know what to do when trying to import `.ts` files.

meanwhile, if you import a `.js` file, TS will (usually) know to look for `.d.ts` file next to it.

test/basic.test.cjs
hchiam: @argyleink - i got the tests to pass locally after `npm run build; npm run bundle;` (it took a while, but stackblitz ran out of memory in my browser) by changing `../` to `./`: 

```suggestion
test('Should produce typings files', async t => {
  t.assert(fs.existsSync('./dist/open-props.d.ts'))
  t.assert(fs.existsSync('./dist/props.sizes.d.ts'))
```
if i understand correctly, from ava's point of view, it's searching for files from top-level folder, not from the point of view of the test file

hchiam, line 58: @argyleink - i got the tests to pass by changing `../` to `./` - after running `npm run build; npm run bundle;` locally (it took a while, but stackblitz ran out of memory in my browser), since that's what's done in [tests.yml](https://github.com/argyleink/open-props/blob/main/.github/workflows/tests.yml#L12): 

```suggestion

test('Should produce typings files', async t => {
  t.assert(fs.existsSync('./dist/open-props.d.ts'))
  t.assert(fs.existsSync('./dist/props.sizes.d.ts'))
```
if i understand correctly, from ava's point of view, it's searching for files from top-level folder, not from the point of view of the test file

build/props.js
```diff
@@ -82,6 +84,32 @@ const figmatokensSYNC = { 'open-props': { ...figmatokens } }
...
+const ES = fs.createWriteStream('../dist/open-props.module.js')
+ES.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
+
+const CJS = fs.createWriteStream('../dist/open-props.cjs')
+CJS.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+// const UMD = fs.createWriteStream('../dist/open-props.umd.js')
+// UMD.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+const TS = fs.createWriteStream('../dist/open-props.ts')
```
argyleink, line 102:
```suggestion
const TS = fs.createWriteStream('../src/open-props.ts')
```

build/props.js
```diff
@@ -82,6 +84,32 @@ const figmatokensSYNC = { 'open-props': { ...figmatokens } }
 const FigmaTokensSync = fs.createWriteStream('../open-props.figma-tokens.sync.json')
 FigmaTokensSync.end(JSON.stringify(figmatokensSYNC, null, 2))
 
+if (!fs.existsSync('../dist'))
+  fs.mkdirSync('../dist')
```
mayank99, line 88: this fails on windows because of `/`.

this works:
```suggestion
const distPath = path.join(fileURLToPath(import.meta.url), '..', 'dist')

if (!fs.existsSync(distPath))
  fs.mkdirSync(distPath)
```

needs these imports:
```js
import path from 'node:path';
import { fileURLToPath } from 'node:url';
```

mayank99: actually i cannot get it to fail anymore. it's really weird.

argyleink:
super glad y'all're makin suggestions so we can collab on this PR 🙂 
i'm also at home with kids hehe, they came home right when i was accepting the first round of suggestions

build/props.js
```diff
@@ -82,6 +84,32 @@ const figmatokensSYNC = { 'open-props': { ...figmatokens } }
...
+// UMD.end(`module.exports = ${JSON.stringify(toObject(), null, 2)}`)
+
+const TS = fs.createWriteStream('../src/open-props.ts')
+TS.end(`export default ${JSON.stringify(toObject(), null, 2)}`)
+
+const TSD = fs.createWriteStream('../dist/open-props.d.ts')
+TSD.end(`declare const OpenProps: ${JSON.stringify(toTypes(), null, 2).replaceAll(',',';')};\nexport default OpenProps;`)
+
+preparedTypes().forEach(({filename, json}) => {
+  let bundle = fs.createWriteStream('../src/props.'+filename+'.d.ts');
```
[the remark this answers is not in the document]
argyleink: lemme give that gitignore an update yep, should be hiding other build files too

mayank99:
yo, you said tomorrow. i was just in the process of testing this with `npm pack` + local apps lol

no worries, i'll test with actual 1.5.4

argyleink:
ah, sorry! i did. it just seemed ready and done.

i updated https://stackblitz.com/edit/react-vanillaextract-openprops to latest and typeahead was there. how'd it go for you? all good? new gitignore got em all?

PR #338, opened 2023-01-24
typescript part 2: electric boogaloo

### LINKED ISSUE #337: issues with typings
Found a few issues in typings when testing 1.5.4:
1. all the primitive types are quoted, i.e. `"string"` and `"number"` instead of `string` and `number`. see [`open-props.d.ts`](https://unpkg.com/browse/open-props@1.5.4/dist/open-props.d.ts).
2. [`props.animations.js`](https://unpkg.com/browse/open-props@1.5.4/src/props.animations.js) only contains kebab-case properties but [`props.animations.d.ts`](https://unpkg.com/browse/open-props@1.5.4/src/props.animations.d.ts) contains both kebab and camel case. same problem applies to all files in `src/` directory
3. [`props.colors.js`](https://unpkg.com/browse/open-props@1.5.4/src/props.colors.js) contains individual and combined objects for colors. [`props.colors.d.ts`](https://unpkg.com/browse/open-props@1.5.4/src/props.colors.d.ts) only contains combined.

all of these could be fixed by using `tsc` instead of generating types by hand 🤔

mayank99:
replaced custom scripts with `tsc` for more accurate types. fixes #337


package.json
```diff
@@ -178,13 +178,14 @@
     "defaults"
   ],
   "scripts": {
-    "build": "concurrently npm:gen:op npm:gen:shadowdom",
+    "build": "\"concurrently npm:gen:op npm:gen:shadowdom\" && npm run gen:types",
```
mayank99, line 181: i'm trying to make `tsc` run after concurrently is finished. it works fine locally (on windows) but keeps failing in CI 😕 maybe i'm missing something obvious

argyleink (approved):
rad

PR #347, opened 2023-02-11
postcss-preset-env v8.0.1

romainmenke:
supersedes : https://github.com/argyleink/open-props/pull/234

unminified diff can be viewed here : https://github.com/romainmenke/open-props/pull/1/files

`postcss-combine-duplicated-selectors` seems to be causing a lot of problems :/
It is breaking @keyframes for example.
But is is also causing large diffs between `postcss-preset-env` version 6.7.0 and 8.0.1.
It is "seeing" different CSS at different times because PostCSS 8 has a different visitor API than PostCSS 7.

It might be best to first remove this dependency on main.
This will fix some bugs like with `@keyframes`.
But it might also require some refactors in imports.

Currently this dependency makes it possible to reference `html` everywhere and then concat later. But this changes the order of CSS in radical ways :/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "open-props",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "open-props",
-      "version": "1.5.4",
+      "version": "1.5.5",
       "license": "MIT",
       "devDependencies": {
         "ava": "^3.15.0",
@@ -15,11 +15,10 @@
         "cssnano": "^5.1.10",
         "json": "^11.0.0",
         "open-color": "^1.9.1",
-        "postcss": "^8.3.9",
+        "postcss": "^8.4.21",
         "postcss-cli": "^8.3.1",
-        "postcss-combine-duplicated-selectors": "^10.0.3",
-        "postcss-import": "^14.0.2",
-        "postcss-preset-env": "6.7.x",
+        "postcss-import": "^15.1.0",
+        "postcss-preset-env": "^8.0.1",
         "typescript": "^4.9.4"
       }
     },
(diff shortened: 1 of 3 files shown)


argyleink:
is `postcss-combine-duplicated-selectors` still misbehaving? i agree it's normally not advised, but for the bundling of props it's great.

romainmenke:
> is postcss-combine-duplicated-selectors still misbehaving? i agree it's normally not advised, but for the bundling of props it's great.

yup, maybe better to create a dedicated plugin that does exactly what you need it to do for `open-props`?

argyleink:
> > is postcss-combine-duplicated-selectors still misbehaving? i agree it's normally not advised, but for the bundling of props it's great.
> 
> yup, maybe better to create a dedicated plugin that does exactly what you need it to do for `open-props`?

merged!

romainmenke:
nice!

A lot has happened since originally opening this pull request and I noticed that we removed a feature somewhere along the way that you depend on here. Transpiling of media range syntax in `@custom-media` is no longer supported. But I think we should bring this back.

I will modernise this pull request after patching that issue.

romainmenke:
@argyleink This is ready for another look.
I've updated the diff preview in the original post at the top, but another link here : https://github.com/romainmenke/open-props/pull/1/commits/ed93c5ed490fc8b494ae86800f1a61d6d11983b3

Summary :
- some white space is different (should largely go away when minifying which I disabled for the preview)
- `;` are now added by the nesting plugin, this fixed some issues for some users
- nesting plugins works slightly differently which seems to result in fewer useless/empty rules
- media range syntax plugin now emits `px` fractions
- order of selector bits in a single compound selector can be different but this will never have a different outcome

argyleink:
I added this back in to `postcss.config.cjs` to remove the `.focus-visible` or `.focus-within` conveniences, and then disabled `:has()` so it just passes through.
```js
'has-pseudo-class': false,
'focus-visible-pseudo-class': false,
'focus-within-pseudo-class': false,
```
pending I do a deeper comparison, this PR is lookin good.

I also need to update the docsite in `/docsite`. Will test out these updates there too and get back to this PR with findings.

Thanks Romain!

romainmenke:
Definitely fine to disable those explicitly but this isn't needed anymore as of version 8.
Any plugin that also requires some client-side JS script is now [disabled by default](https://github.com/csstools/postcss-plugins/tree/main/plugin-packs/postcss-preset-env#enableclientsidepolyfills).

We found that too many users had these plugins enabled without knowing this and without loading the client-side JS scripts :)

argyleink:
closing. can be reopened if new interest sparks.

Issue #354, opened 2023-03-12
Add link to this color helper in docsite 


argyleink:
https://props-color-finder.netlify.app

kaf-lamed-beyt:
@argyleink I can take this, if you'd love me to.

I'm thinking of creating a section just after "Tooling", below — for projects built with open-props, more like a showcase of some sort.

https://github.com/argyleink/open-props/blob/58a52be43251681e23678d4b93df898a569b51d3/docsite/index.html#L3491

Or I can just add it to the tooling div without any hassle. Please let me know which one sounds good.

argyleink:
hey thanks! yeah, i think that's a great spot 👍🏻 

add a new div and `<dt>Showcase</dt>`, we'll start piling showcase links in there 🚀

kaf-lamed-beyt:
Awesome!! 

I'll do that now. 🚀

PR #355, opened 2023-03-17
chore: include color helper

kaf-lamed-beyt:
#354 

Hi @argyleink, kindly take a look.

### DIFF
--- a/docsite/index.css
+++ b/docsite/index.css
@@ -156,7 +156,7 @@ footer {
 
     & > div {
       display: grid;
-      place-self: center;
+      place-self: start center;
       place-items: center;
       flex-basis: var(--size-content-1);
       gap: var(--size-2);
@@ -167,10 +167,6 @@ footer {
       text-transform: uppercase;
     }
   }
-}
-
-footer dl > div {
-  place-self: start center;
 }
 
 section.hero {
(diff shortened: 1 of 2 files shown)


argyleink (changes requested):
Looks great!

Can you update a style too, so it aligns better on desktop? Here's what needs changed.

```css
footer dl > div {
  place-self: start center;
}
```

kaf-lamed-beyt:
> Can you update a style too, so it aligns better on desktop? Here's what needs changed.

All done.

Learned something new today, because this is the first time I'm seeing this particular CSS property &mdash; `place-self`

Thank you for the opportunity, @argyleink!

docsite/index.css
```diff
@@ -169,6 +169,10 @@ footer {
   }
 }
 
+footer dl > div {
+  place-self: start center;
+}
+
```
argyleink, line 175: can you change it [here instead](https://github.com/argyleink/open-props/blob/main/docsite/index.css#L159) please =)

argyleink (changes requested):

docsite/index.css
kaf-lamed-beyt: I had a guess about this too. I'll change it now.

kaf-lamed-beyt: done

argyleink (approved):
ty! 💯

kaf-lamed-beyt:
🫡

PR #365, opened 2023-04-21
Fix accessibility issues on theme toggle button

mobalti:
I fixed the attributes on the theme toggle button so that screen readers can recognize the current state of the light and dark theme.


argyleink (approved):
lgtm 🙏🏻

PR #368, opened 2023-05-07
Text cursor on text inputs

### LINKED ISSUE #366: form inputs shouldnt be cursor pointer
they should be cursor text

rienheuver:
Fixes #366

Not sure if you'd need to keep the `touch-action` and `-webkit-tap-highlight-color` on these excluded inputs. If so, that part should perhaps be extracted to a new block. Let me know if you'd prefer that

### DIFF
--- a/src/extra/normalize.src.css
+++ b/src/extra/normalize.src.css
@@ -76,8 +76,11 @@
   }
 }
 
-:where(a[href], area, button, input, label[for], select, summary, textarea, [tabindex]:not([tabindex*="-"])) {
+:where(a[href], area, button, input:not([type="text"], [type="email"], [type="number"], [type="password"], [type=""], [type="tel"], [type="url"]), label[for], select, summary, [tabindex]:not([tabindex*="-"])) {
   cursor: pointer;
+}
+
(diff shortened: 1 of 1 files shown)


argyleink (changes requested):
yeah, i think a new block is a good idea 👍🏻 i do like getting rid of those highlights, they're uncanny. maybe i should just wipe them entirely/globally and not via a selector 🤔 

also, `input[type="date"]` should be omitted from the cursor pointer too yeah?

thanks for this! gonna be a nice subtle change 🙂

rienheuver:
I separated the highlights into another block. I'll leave wiping them globally upto you/another PR and consider it out of scope for this.

As for the `[type="date"]: I only targeted the ones that my browser (Chrome) gives `cursor: text` by default and I arrived at that list by going through the ones on this page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

I'm not sure why `date` doesn't get the text-cursor there, it seems like a limbo case. I can follow adding it, but leaving it out makes equally much sense to me I guess. Either way, if we'll include `date`, than we should include `datetime-local`, `month`, `time` and `week` as well.

Let me know what you think :smile:

argyleink (approved):
thanks!

PR #370, opened 2023-05-08
add relative sizes to OP

### LINKED ISSUE #336: Add relative sizes
@therien brought this up 👍🏻 

```css
html {
  --size-relative-0: .25ch;
  --size-relative-1: .5ch;
  --size-relative-2: 1ch;
  etc…
}
```

would be nice to have these as offerings within the lib. well named, purpose built props for relative sizing of elements and components.

would pair with:

```css
--size-{000-15}
--size-fluid-{1-10}
--size-content-{1-3}
--size-header-{1-3}
```

my main resistance to this so far has been that there's many types of relative units, all with great use cases, and it wouldnt be a good idea to create a set of props for each of those length types. aka, **this doesnt scale well**. but, i personally use `ch` for 90% of it all, and having them would be nice to replace the `ch` units used throughout the `normalize/buttons` files with a prop.

PaoloGiraudi:
Fixes #336 

Not sure if the range `.25ch` --> `12ch` is enough or too much


docsite/index.html
```diff
@@ -2974,6 +2975,58 @@ <h5 style="inline-size: var(--size-header-3)">Carrot cake gummi bears sweet cara
         </div>
       </div>
 
+      <div class="block-wrap">
+        <div>
+          <h5>Relative Sizes</h5>
```
argyleink, line 2980: this title is the same as the regular sizes, which makes me think the regular sizes title should change to either "Document Relative Sizes" or just "Sizes".

docsite/index.css
```diff
@@ -220,13 +220,13 @@ section.hero {
 
   & > * {
     background: var(--gradient-1) fixed;
-    background-size: 12ch 7ch;
+    background-size: var(--size-relative-15) var(--size-relative-12);
```
argyleink, line 223: thanks for dogfooding / backporting these length updates!

docsite/index.html
```diff
@@ -2974,6 +2975,58 @@ <h5 style="inline-size: var(--size-header-3)">Carrot cake gummi bears sweet cara
...
+              --size-relative-11: 6ch;
+              --size-relative-12: 7ch;
+              --size-relative-13: 8ch;
+              --size-relative-14: 10ch;
+              --size-relative-15: 12ch;
+            </code></pre>
+          </div>
+        </div>
+        <div>
+          <h5>Usage Sample</h5>
```
argyleink, line 3003: be cool if this demo area had a "font size picker" select element so you could see the sizes adapt to a given font size. really help set the feature in. thoughts? it'd be kinda like how line-height and font-size have on.

src/props.sizes.js
```diff
@@ -43,4 +43,21 @@ export default {
...
+  '--size-relative-6': '1.75ch',
+  '--size-relative-7': '2ch',
+  '--size-relative-8': '3ch',
+  '--size-relative-9': '4ch',
+  '--size-relative-10': '5ch',
+  '--size-relative-11': '6ch',
+  '--size-relative-12': '7ch',
+  '--size-relative-13': '8ch',
+  '--size-relative-14': '10ch',
+  '--size-relative-15': '12ch',
```
argyleink, line 62: I wonder if this list should be the exact same as the rem sizes but just swapped to the `ch` unit? it'd make interchanging between them easier AND the mental model of the scale easier. We'd also gain the 2 negative values. So like, less to manage and more consistent? thoughts?

argyleink (changes requested):
thank you for this! 🙏🏻  
not sure how it skipped my radar, i try and stay on these PRs. 

i've left a few comments, nothing critical.

the one thing i'm unsure of, is if these should ship in the main bundle or as an optional import? have thoughts on this? my first thought was to ship as an additional import, but then again, the colors alone are what really bloat the package not the sizes, and in a future release i'll be heavily cutting back the colors shipped. so I dunno! i'm on the fence.

PaoloGiraudi:
@argyleink A couple notes: 
 - In my opinion, these variables should be part of the main bundle. Having `ch` and `rem` units by default will give users a great toolkit to build UI. 
 - By using the same values as for `rem`, we don't have [this specific size](https://github.com/argyleink/open-props/pull/370/files#diff-c46b9c0fde3e24499651d4726fdf2b5368926cd9e62196c9f3976dca73b9c4e9R48) used in `buttons`.

docsite/index.css
```diff
@@ -220,13 +220,13 @@ section.hero {
 
   & > * {
     background: var(--gradient-1) fixed;
-    background-size: 12ch 7ch;
+    background-size: var(--size-relative-13) var(--size-relative-11);
```
argyleink, line 223: not exactly the same, but still works 👍🏻

argyleink (approved):
rad work!

docsite/index.html
argyleink: you did perfect 👍🏻

argyleink:
available in v1.5.9 🎉

PR #399, opened 2023-08-29
adds inner-shadow-highlight variable to inner shadow 1-4

argyleink:
Screenshot 2023-08-29 at 3 54 26 PM

### DIFF
--- a/src/props.shadows.css
+++ b/src/props.shadows.css
@@ -3,6 +3,7 @@
 :where(html) {
   --shadow-color: 220 3% 15%;
   --shadow-strength: 1%;
+  --inner-shadow-highlight: inset 0 -.5px 0 0 #fff2, inset 0 .5px 0 0 #0007;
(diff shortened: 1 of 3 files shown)


argyleink:
tests are failing because I added a new prop

ezzatron:
@argyleink Just a heads up that the inner shadow highlight looks a bit... weird on light themes.

With `--inner-shadow-highlight` it looks kind of like there's a dark 1px border along the top edge of the inset:

Screenshot 2023-10-24 at 20 56 32

Without `--inner-shadow-highlight`:

Screenshot 2023-10-24 at 20 56 21

(Note that this component deliberately only uses the inset shadow on the top and bottom)

As usual, it's easy to override, but I thought I'd bring it up in case it was just an oversight. Cheers!

argyleink:
ewww that's gross agree, lemme fix!

argyleink:
fixed in `v1.6.9` 🙏🏻

PR #400, opened 2023-09-01
fix breaking centered text-alignment for td, th

kalligator:
I noticed that :where(td,th) has text-align: left set, which messes things up when td or th has align="center" in their html tag. 

I fixed this issue, but I think it would be better to remove the alignment from :where(td,th) altogether, because it might also cause problems for right-to-left users.

### DIFF
--- a/src/extra/normalize.src.css
+++ b/src/extra/normalize.src.css
@@ -346,6 +346,11 @@
   padding: var(--size-2);
 }
 
+td:where([align="center"]),
+th:where([align="center"]) {
+	text-align: center;
+}
+
 :where(thead) {
   border-collapse: collapse;
 }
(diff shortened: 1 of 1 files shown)


argyleink:
Thank you for filing this 🙂 

I'm surprised I committed a physical property for text alignment, I assumed I would have committed `text-align: start`.

What if we did this:

```css
:where(td:not([align]), th:not([align])) {
  text-align: start;
}
```

this way
- `start` is always the default alignment (wont break for right to left users) 
- will only apply if the `align` attribute is absent from the cell
- the `:not()` is within the `:where()` as to not increase specificity and be easy to override

kalligator:
That sounds a lot better indeed!

src/extra/normalize.src.css
```diff
@@ -346,6 +346,11 @@
   padding: var(--size-2);
 }
 
+td:where([align="center"]),
+th:where([align="center"]) {
```
argyleink, line 350:
```suggestion
:where(:is(td,th):not([align])) {
```

PR #401, opened 2023-09-02
(feat) Better support for row colors in tables

SanderElias:
setting a color on the `td` prevents easy zebra striping. this commit moves the color to the table and the headerfoorter so it becomes easier to override the `tr` background color

### DIFF
--- a/src/extra/normalize.src.css
+++ b/src/extra/normalize.src.css
@@ -296,7 +296,7 @@
 :where(table) {
   width: fit-content;
   border: 1px solid var(--surface-2);
-  background: var(--surface-2);
+  background: var(--surface-1);
   border-radius: var(--radius-3);
 
   --nice-inner-radius: calc(var(--radius-3) - 2px);
@@ -336,7 +336,6 @@
 }
 
 :where(td) {
-  background: var(--surface-1);
   max-inline-size: var(--size-content-2);
   text-wrap: pretty;
 }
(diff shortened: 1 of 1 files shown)


argyleink:
I tested out the changes and the updated table styles aren't 1:1 with the current:
- border around the table should match
- gaps between cells are missing
- (screenshot has me hovering that darker row, my bad)

Screenshot 2023-09-03 12 01 32 PM

I'm also curious why targeting the `td` is a hassle? That worked great for me, and maintains the cell gaps around each td (which is why they were colored all this time 😅 .

```css
tbody > tr:nth-child(even) td {
  background: var(--surface-3);
}
```

Screenshot 2023-09-03 12 04 01 PM

argyleink:
I tested out the changes and the updated table styles aren't 1:1 with the current:
- border around the table should match
- gaps between cells are missing
- (screenshot has me hovering that darker row, my bad)

Screenshot 2023-09-03 12 01 32 PM

I'm also curious why targeting the `td` is a hassle? That worked great for me, and maintains the cell gaps around each td (which is why they were colored all this time 😅 .

```css
tbody > tr:nth-child(even) td {
  background: var(--surface-3);
}
```

Screenshot 2023-09-03 12 04 01 PM

SanderElias:
🤦
```css
tbody > tr:nth-child(even) td {
  background: var(--surface-3);
}
```

That is a much better solution! Not sure where my mind was!

PR #413, opened 2023-09-20
fix typo

romainmenke:
Without the `;` the selectors and rules after the last custom media technically become part of the at rule prelude.


argyleink (approved):
nice catch, ty!

Issue #248, opened 2022-07-15
Add `--ease-anticipate-*` easings


argyleink:
these would be complimentary to `--ease-elastic-*` in that elastic easing ends with a bounce and the anticipate easings would start with a bounce. we have `--ease-squish-*` which bounces both sides also. so adding these easings would give folks that disney type [anticipatory animation](https://www.joshuanava.biz/disney-animation/anticipation.html) start. 

this task is to add 5 more easings that should be very similar to elastic

todo:
- [ ] add 5 new props
- [ ] update docsite with new props

argyleink:
a strong proposal by Jhey is to name it `--ease-elastic-in-*` 👍🏻  the only bummer is it might mean renaming the other elastic easings and creating a breaking version. i think i'm down with the tradeoff tho, cuz this looks nice:

```css
/* today */
--ease-elastic-3
--ease-anticipate-3 (to do)
--ease-squish-3

/* proposed breaking change */
--ease-elastic-out-3
--ease-elastic-in-3
--ease-elastic-in-out-3
```

argyleink:
`--ease-windup-*` is another contender

argyleink:
```css
--ease-out-elastic-3
--ease-in-elastic-3
--ease-in-out-elastic-3
```

621alexl:
Working on this now for TS Open.

PR #419, opened 2023-09-29
Added --ease-anticipate-* easings

621alexl:
**breaking change**

Added --ease-elastic-in, renamed --ease-elastic-* and --ease-squish-* to --ease-elastic-out-* and --ease-elastic-in-out-* (rename as per [Adam's comment](https://github.com/argyleink/open-props/issues/248#issuecomment-1185724121)). Corresponding changes to docsite were made (adding --ease-elastic-in-* demo box and refactored for the rename).

P.S: First time OSS contributor and new to CSS, sorry it there are any issues w/ my Bezier functions or anything else.


argyleink (changes requested):
very thorough! thank you!

you looked around and found the ancillary areas of usage and examples and made sure it was covered 🤘🏻 

I found one type and made a suggestion about it.

Other than that, the tests need updated because there's new props!
1. https://github.com/argyleink/open-props/blob/main/test/basic.test.cjs#L7 is now 1622
2. https://github.com/argyleink/open-props/blob/main/test/basic.test.cjs#L27 is now 424

<br>

> breaking change

I fully agree this is a breaking change! I could make a v2 branch and we can adjust this PR to push to that (at which point I can start hacking on v2 changes more), but, we could also do a minor version change to 1.6.0 and put a note in the docsite about the name change? or we could even keep both names around (old squish and elastic) ones, and provide a deprecation notice. thoughts on this? in terms of a breaking change, it's pretty minor, but breaking nonetheless.

621alexl:
Accepted your changes Adam! I also made another commit to adjust the tests. Thanks for catching the typo. As for your discussion of making a v2 branch, I feel it would be nice to give users this additional prop. Why don't we go with your send proposal of making a change to 1.6.0, aliasing `--ease-squish-*` and `--ease-elastic-*` to avoid breaking anything.

Issue #406, opened 2023-09-19
Adapter for Uno


argyleink:
https://unocss.dev/presets/community

argyleink:
thanks jeff! great thoughts. I'd originally imagined a sub-export yeah, something like `import "open-props/presets/uno";` that was a file built by a new build script like [this to-object one](https://github.com/argyleink/open-props/blob/main/build/to-object.js), it'd be like, `to-preset-uno.js`. sounds like we're aligned in this! i think you'll find the workflow for calling that build script and testing the output a quick feedback loop. sorry for some of the complexity showing up in [`props.js`](https://github.com/argyleink/open-props/blob/main/build/props.js#L23) but i think you'll be able to add your build task there easy enough. dont hesitate to ask me questions, i'll make sure you're moving along nicely!

for the `devDep` addition, i'm not 100% sure on that either. i kinda feel like, for now at least, we just export the file and let authors manage their uno installation/dep, and this package just offers the preset. but i dunno, maybe uno has opinions about it.

jeffposnick:
I'm learning a bit more about how `uno` learns about style tokens (as [part of a `theme`](https://unocss.dev/config/theme)) and then applies those tokens ([via `rules`](https://unocss.dev/config/theme#usage-in-rules)).

This changes how I'm going to structure the exported configuration a bit, but it makes sense since `uno` is library whose focus is actually applying style definitions via rules.

bjorntheart:
@argyleink following on from our discussion on https://github.com/argyleink/open-props/pull/423, I wanted to make sure of a couple of things first.

In my over simplified understanding about the work @jeffposnick started on https://github.com/argyleink/open-props/pull/423, the `/^op-color-(.*)$/`  _rule_ converts an OP prop into an atomic CSS class using the token's value `var(--violet-4)`?

**Example**

> if the OP was `--violet-4` the class ends up begin 

```css
op-color-violet-4 {
  color: #9775fa;
}
```

More _rules_ in the preset would extrapolate OP custom properties into its respective atomic CSS classes.

## Atomic CSS Class Prefix

What do you think of having the atomic class prefix be configurable via the preset, but we also provide a default to `op`?

**Example**

```js
import { defineConfig } from 'unocss'
import openPropsPreset from "open-props/presets/uno";

export default defineConfig({
  presets: [
    openPropsPreset({
      prefix: 'ops', // default value is 'op'
    }),
  ],
})
``` 
[...]

PR #423, opened 2023-09-29
[Draft]: uno preset

jeffposnick:
This is a work-in-progress PR, which fixes https://github.com/argyleink/open-props/issues/406

I learned a lot about how to translate CSS variables into something that could be effectively applied as a style by a framework like https://unocss.dev/

I need to take a look at what other themes and rules it would make sense to define as part of this preset.

As it stands, you can apply styles that use the font and color variables via class names like `op-font-sans` or `op-color-choco-4`.

### DIFF
--- a/src/presets/unocss/index.js
+++ b/src/presets/unocss/index.js
@@ -6,8 +6,8 @@ export const openPropsPreset = {
   rules: [
     [
       /^op-color-(.*)$/,
-      ([, c], { theme }) => {
-        const normalizedVariable = `--${c}`;
+      ([, token], { theme }) => {
+        const normalizedVariable = `--${token}`;
         if (theme["colors"][normalizedVariable]) {
           return { color: theme["colors"][normalizedVariable] };
         }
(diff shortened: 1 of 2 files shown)


src/presets/unocss/index.js
```diff
@@ -0,0 +1,30 @@
+import Colors from "../../../src/props.colors.js";
+import Fonts from "../../../src/props.fonts.js";
+
+export const openPropsPreset = {
```
argyleink, line 4: sick, do the work in the module export with the js source files, very nice! much better than a build file's output

bjorntheart:
@jeffposnick Are you still working on this? I Would love to help. I'm doing something similar, but would rather focus my efforts here and contribute.

jeffposnick:
Sorry, no, I haven't been able to devote the time to wrapping this up. Feel free to take over and reuse anything that's useful!

bjorntheart:
Thanks @jeffposnick.

@argyleink I'm taking over from @jeffposnick on this one. What's the best approach to go about this, create a brand new PR, and mention this one for context?

argyleink:
I think you could PR his PR, so then he gets contributor status when it all merges? But maybe that puts a bit of work on @jeffposnick that they don't want? I can also help make @jeffposnick a contributor in an easy way after a pr merges that resolves the feature. 

I'm good either way! Just wanna make sure folks get their cred

bjorntheart:
> I'm good either way! Just wanna make sure folks get their cred

Agreed!

@argyleink Think I'm going to make a new PR for this one ;-)

jeffposnick:
I'm good either way! Whatever is easiest for you.

PR #428, opened 2023-10-14
fix referencing of elastic easing

waspeer:


waspeer:
I've added a test as well that checks if all references are valid.

argyleink (approved):
excellent catch, ty!

argyleink:
fixed in v1.6.6

PR #435, opened 2023-11-13
fix(props.js): prefix mask individual props file names as all others

cbontems:
I just noticed the 2 individual mask files where not prefixed on a custom prefix build.
This PR simply adds the prefix to these 2 files like for others.


argyleink (approved):
awesome, thank you 🙂

PR #436, opened 2023-11-16
Chore: add `width: fit-content` to a tag

jacobthesheep:
Should be default style

### DIFF
--- a/src/extra/normalize.src.css
+++ b/src/extra/normalize.src.css
@@ -91,7 +91,7 @@
   margin-inline: calc(var(--size-1) * -1);
   padding-block: var(--size-1);
   margin-block: calc(var(--size-1) * -1);
-  width: fit-content;
+  max-inline-size: fit-content;
 
   &:where([href]) {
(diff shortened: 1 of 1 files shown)


src/extra/normalize.src.css
```diff
@@ -91,6 +91,7 @@
   margin-inline: calc(var(--size-1) * -1);
   padding-block: var(--size-1);
   margin-block: calc(var(--size-1) * -1);
+  width: fit-content;
```
argyleink, line 94:
```suggestion
  max-inline-size: fit-content;
```

argyleink (changes requested):
remove all the formatter changes and update to `max-inline-size` and I think this is g2g!

argyleink:
need to revert this, caused unexpected side effects for folks wrapping cards or groups of content with a link. i think this needs to go back on the developer to fix when the links are stretching beyond what they need. aka, links needs to be more naive to sizing in normalize, we'll set max-inline-size on links when they're being stretched in undesirable ways.

Issue #443, opened 2023-12-05
Extend --HDcolor custom MQ with color-gamut


cbontems:
My MacBook pro has a p3 capable display, but returns false to the --HDcolor media query.

I ended up extending this media query this way:
`@custom-media --HDcolor (dynamic-range: high) or (color-gamut: p3);`

This way, I can benefit from these gorgeous colors.
I am probably not the only one in this situation, so if this sounds like a good update, I would be happy to submit a PR.

cbontems:
Hi Adam!
Well, I am using Chrome, but this is not Chrome specific. Safari is the only one to return true. Chrome / Brave / Arc / Firefox all return false.

When using (color-gamut: p3), only Firefox keeps denying me the right to display nice colors...

I use a quite old but faithful 2018 15" mbp.

By the way, Chrome dev tools allow to emulate the color-gamut MQ, but not the dynamic-range MQ. Is there a specific reason for that?

PR #444, opened 2023-12-05
fix(customMQ): extend --HDcolor media query with color-gamut check

cbontems:
Fix issue #443 
Extends the --HDcolor custom media query the check for either a dynamic-range: high or color-gamut: p3

### DIFF
--- a/docsite/index.html
+++ b/docsite/index.html
@@ -3352,7 +3352,7 @@
               @custom-media --pointer  (hover) and (pointer: coarse);
               @custom-media --mouse    (hover) and (pointer: fine);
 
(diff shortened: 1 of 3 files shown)


cbontems:
Oh, by the way, I often use this one:
  `--forcedColors: (forced-colors: active)`
It would be nice to have it in the default Open Props set.
I can add it too.

argyleink:
> Oh, by the way, I often use this one: `--forcedColors: (forced-colors: active)` It would be nice to have it in the default Open Props set. I can add it too.

oh yes! how'd i miss it. ty ty, please add if you are down 🙂

PR #445, opened 2023-12-07
feat(custom-media): add --forcedColors to the custom media queries list

cbontems:
Oh, I missed the timing to slide this one in the previous PR, so here it is in a new one.

Adds `@custom-media --forcedColors  (forced-colors: active);` to the custom MQ list.


argyleink (changes requested):
just needs added to the docsite as well, then this is g2g!

argyleink (approved):
rad! ty, this is great 👍🏻

cbontems:
My pleasure

PR #450, opened 2023-12-15
Update non core dependencies

nix6839:
These updates do not significantly affect the bundled code.
Upon checking, it only affects the order of CSS declarations.


argyleink:
thanks for contributing but efforts to update deps and risk ordering changes in v2

PR #456, opened 2023-12-28
make figma-tokens plugin work again

### LINKED ISSUE #455: value and type -> $value and $type causes issues in current version of figma-tokens
while updating [my notes and videos of setting up op in figma-tokens](https://github.com/hchiam/figma-tokens-sync-example), i noticed tokens aren't properly generated in the latest figma-tokens plugin when i try to use either of the following URLs: 
- https://unpkg.com/open-props@1.6.16/open-props.figma-tokens.json or 
- https://unpkg.com/open-props@1.6.16/open-props.figma-tokens.sync.json 

Commit https://github.com/argyleink/open-props/commit/327a70f82aa3aa6c553138ffd1572b3c4fd0ab18 was done to [update design token json to the latest spec syntax](https://github.com/argyleink/open-props/issues/392), but i'm thinking we'll need to undo part of the commit to treat figma tokens differently. 

it seems the [figma-tokens plugin](https://github.com/tokens-studio/figma-plugin) is still expecting `value` and `type` instead of `$value` and `$type` because it keeps trying to recreate `value` and `type` when i hit "Save JSON", but it does so badly: 

figma-tokens tries to convert from $value to value but does it badly

and the output does not look right:

the output has $value and $type as individual values, with everything under Other

hchiam:
undo the `build/to-figmatokens.js` part of 327a70f (#392) 
to fix #455 and make op tokens work again in [figma-tokens plugin](https://github.com/tokens-studio/figma-plugin)

output looks correct again: (compare with the 2 screenshots in https://github.com/argyleink/open-props/issues/455#issue-2057942113) 

value and type instead of $value and $type

open props color tokens show colors again and are no longer just saying $value and $type

### DIFF
--- a/build/to-figmatokens.js
+++ b/build/to-figmatokens.js
@@ -15,25 +15,25 @@ export const toFigmaTokens = props => {
       .map(hueName => hueName.toLowerCase())
     let isColor = colors.some(color => key.includes(color))
     
-    if      (isLength) meta.$type = 'sizing'
-    else if (isBorder) meta.$type = 'borderWidth'
-    else if (isRadius) meta.$type = 'borderRadius'
-    else if (isShadow) meta.$type = 'boxShadow'
-    else if (isColor)  meta.$type = 'color'
-    else               meta.$type = 'other'
+    if      (isLength) meta.type = 'sizing'
+    else if (isBorder) meta.type = 'borderWidth'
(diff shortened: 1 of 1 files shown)


argyleink (approved):
ty 🙏🏻

PR #464, opened 2024-01-29
Exporting `./supports.min.css`

SalahAdDin:
The file is listed on the docs page, but when importing it, `node` is unable to find it.


argyleink:
node can find it if you don't use the exports aliases, if you instead point the full path to the file in node_modules

`'open-props/src/props.supports.css'`

the reason it's not listed in the exports of package.json is because it's generally shunned to target specific browsers and I don't want to promote the authoring of specific code. i am ok supporting folks who need it, but not promoting it. so that's why it's there but not made as easy as the rest of the imports.

PR #469, opened 2024-02-20
feat(media): add inverted colors custom media query

cbontems:
Hi Adam,  

Browser support in not there yet, except for Safari, and Firefox behind flags, but I suggest we add the inverted colors MQ to the custom media queries list.

Updated:
- docsite
- props.media.css
- props.media.js

[Spec for reference](https://drafts.csswg.org/mediaqueries-5/#inverted)


argyleink (approved):
looks great, thank you!

powerrangers

PR #472, opened 2024-02-23
Fix misconfigured Open Graph metadata

yudai-nkt:
I was testing my link card component with https://open-props.style/, and noticed that the website's Open Graph metadata is not correctly set. I fixed it accordingly.


argyleink (approved):
thank you!

PR #475, opened 2024-02-26
durations.css – correct leap-year duration; introduce olympiad

trych:
Hi there,

just noted an error in the durations. A leap year does not refer to a time span of four years, but to the length of the year where the leap day occurs, i.e. 366 days.

On the other hand, conveniently, there *is* a duration that describes a four-year timespan, an olympiad, which I introduced as a new property.

Hope this makes sense.


argyleink (approved):
rad 🤘🏻 
i bet the tests will need update to account for the addition, will let the action run and verify

argyleink:
oh nice, tests are only for the main bundle, not for extra packs. this is g2g!

PR #476, opened 2024-02-29
Implement clickable color swatches to copy color props to clipboard

### LINKED ISSUE #467: [enhancement] Click color swatch to copy hex code to clipboard?
I would first like to say that I absolutely love Open-Props and I think it's such a simple yet effective way to have consistently good styling across my projects.

That being said, I often find myself wanting to just steal a color or two for something really small like submitting a component on [UIverse.io](https://uiverse.io) (where I can't import external stylesheets) or even using a color for a Google Slides presentation.

I think it would be useful to be able to easily grab a hex code from the color swatches in the colors section of the site. Though I do realize that it could be confusing and defeat the purpose of having the props available as a library.

Here's a simulation of my idea:
GIF of clicking on a color swatch and copying the color code into Google Slides

If you think this would be a useful addition to the site I would be happy to submit a PR implementing this feature. Thanks!

Brian-Pob:
Closes #467


docsite/index.css
```diff
@@ -1245,13 +1271,18 @@ pre[class*="language-"] {
 
 .count-em {
 
-  &::before {
-  content: '';
-  counter-reset: basic;
-  position: absolute;
-}
+  &:not(.tooltip, li):before {
```
argyleink, line 1274:
```suggestion
  &:not(.tooltip, li)::before {
```

docsite/index.css
```diff
@@ -1245,13 +1271,18 @@ pre[class*="language-"] {
...
+    content: '';
+    counter-reset: basic -1;
+    position: absolute;
+  }
+
+  & > *:not(.tooltip, li):before {
```
argyleink, line 1280:
```suggestion
  & > *:not(.tooltip, li)::before {
```

docsite/index.css
```diff
@@ -1245,13 +1271,18 @@ pre[class*="language-"] {
...
+    counter-increment: basic;
+    content: counter(basic);
+  }
 
-  & > *::before {
+  & button:before {
```
argyleink, line 1285:
```suggestion
  & button::before {
```

docsite/js/color-copy.js
```diff
@@ -0,0 +1,184 @@
...
+
+    if (cb_hsl?.checked) {
+      copiedText = RGBToHSL(r, g, b);
+    }
+  }
+
+  if (copiedText === '')
+    return;
+
+  console.log(copiedText);
```
argyleink, line 133:
```suggestion
```

docsite/js/color-copy.js
```diff
@@ -0,0 +1,184 @@
...
+  navigator.clipboard.writeText(copiedText);
+  return copiedText;
+}
+
+function attachCopyEvent(query, property) {
+  const elements = document.querySelectorAll(query);
+
+  elements.forEach((element) => {
+    element.addEventListener('click', function (e) {
```
argyleink, line 142: this is a good moment to leverage event bubbling, for 1 listener instead of 1 on each swatch!? down to refactor it a bit? shouldnt effect the rest of the code, mostly removes this loop and adds a click guard for clicks on other stuff in the parent. thoughts? 😅

docsite/js/color-copy.js
```diff
@@ -0,0 +1,184 @@
...
+      const copiedText = copyPropertyToClipboard(e.target, property);
+      addTooltip(e.target, copiedText);
+    });
+  });
+}
+
+function selectCopyFormat() {
+  const cb_prop = document.querySelector('#prop');
+  // const cb_color = document.querySelector('#color-code');
```
argyleink, line 151:
```suggestion
```

docsite/index.html
```diff
@@ -1319,311 +1319,350 @@ <h5>Modify Opacity Example</h5>
...
+              <label for="color-code">
+                Color Code <span class="example">Ex: rgb(3, 5, 7)</span>
+              </label>
+            </div>
+          </div>
+        </form>
+
+        <form>
+          <p>Copy format</p>
+          <div class="options">
```
argyleink, line 1344: i dont think this needs a refactor, this is more of a comment: 
I personally would have used a <fieldset> and <legend> for these 2 <forms>, which then would make listening for user inputs only 1 listener on each fieldset and the added bonus you can disable a fieldset which disables the children. slightly simpler to manage i think, but would need some css updates to remove the styling of the html elements.

all in all, what you have here is accessible to keyboard and screen reader, so no refactor needed 👍🏻

docsite/color-copy.css
```diff
@@ -0,0 +1,109 @@
...
+    font-style: italic;
+    color: var(--text-2);
+  }
+
+  [data-theme="light"] & .example {
+    color: var(--gray-7);
+  }
+
+  & input:is(:disabled) {
+
```
argyleink, line 18:
```suggestion
```

argyleink (changes requested):
there's a few nits in there, like removing console logs or commented out stuff. all in all this is super clean!

the color conversions are a bit noisy and there's good color libs, but i think it's totally fine for this to have it's own functions in place. IF you wanted, no pressure here, to pull in say [colorjs](https://colorjs.io) to do the conversion work, it'd also **be cool** to add a couple other color functions to the options: `oklch` and `p3` (rounded 1 or 2 decimal places) as more modern syntax offerings.

nice tooltip solution 👍🏻 
would have been fun to use `[popover]`, but needs a little bit better browser support (it's like about to drop in FF), and I dont think this use case really necessitates a rewrite to use it (`:top-layer`). 

just a few little things and i think this is g2g. thanks a bunch, folks will love it!

docsite/js/color-copy.js
Brian-Pob: I would be down to implement it. I'll brush up on event bubbling and how to implement it properly.

argyleink (approved):
rangers

PR #477, opened 2024-03-04
Fix the broken link for durations.css

ibelem:
The link of https://github.com/argyleink/open-props/blob/main/src/props.durations.css in docsite is broken, PTAL.

### DIFF
--- a/docsite/index.html
+++ b/docsite/index.html
@@ -356,7 +356,7 @@
               </div>
             </dd>
             <dt>Extras</dt>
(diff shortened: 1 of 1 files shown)


argyleink (changes requested):
ah! nice catch, ty!

can you reduce the PR to just the link change tho please? then this is an easy merge 🙂

ibelem:
@argyleink Updated the PR to change the broken link only. PTAL

argyleink (approved):
rad ty!

PR #496, opened 2024-04-18
Drawn Borders

Jothsa:
I added the drawn borders. I went ahead and added an example of using them since I thought people might not necessarily connect getting a hand-drawn effect with applying a border radius. For the tests, I just changed the numbers until they worked 😅


argyleink (approved):
sorry it took so long for me to review. everything looks good!

argyleink:
available in `v1.7.4` 🎉

PR #498, opened 2024-05-01
Add Modern Font Stacks

Jothsa:
Let me know if there's anything else I need to do. I think it might be nice to have an explanation of the benefits of using these props over an external font, but I'm not sure where it would fit. Thanks!


argyleink:
Look at all that tucked into nice little props 👍🏻  
I do wonder if we need to break these modern font stacks into their own import? 

Screenshot 2024-05-02 at 2 31 26 PM

Screenshot 2024-05-02 at 2 32 01 PM
Screenshot 2024-05-02 at 2 32 08 PM

wonder if we should put ^ these behind a summary details so they don't take up so much space? was cool when there were a few, but now there's a lot.

docsite/index.html
```diff
@@ -2114,6 +2114,22 @@ <h4>Font Families</h4>
...
+          --font-monospace-slab-serif: 'Nimbus Mono PS', 'Courier New', monospace;
+          --font-monospace-code: ui-monospace, 'Cascadia Code', 'Source Code Pro', Menlo, Consolas, 'DejaVu Sans Mono', monospace;
+          --font-industrial: Bahnschrift, 'DIN Alternate', 'Franklin Gothic Medium', 'Nimbus Sans Narrow', sans-serif-condensed, sans-serif;
+          --font-rounded-sans: ui-rounded, 'Hiragino Maru Gothic ProN', Quicksand, Comfortaa, Manjari, 'Arial Rounded MT', 'Arial Rounded MT Bold', Calibri, source-sans-pro, sans-serif;
+          --font-slab-serif: Rockwell, 'Rockwell Nova', 'Roboto Slab', 'DejaVu Serif', 'Sitka Small', serif;
+          --font-antique: Superclarendon, 'Bookman Old Style', 'URW Bookman', 'URW Bookman L', 'Georgia Pro', Georgia, serif;
+          --font-didone: Didot, 'Bodoni MT', 'Noto Serif Display', 'URW Palladio L', P052, Sylfaen, serif;
+          --font-handwritten: 'Segoe Print', 'Bradley Hand', Chilanka, TSCu_Comic, casual, cursive;
+          --font-emoji: 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
```
argyleink, line 2132: i'm curious if these need the quotes around them or not?

docsite/index.html
```diff
@@ -2134,6 +2150,96 @@ <h4>Font Families</h4>
               <span>A</span><span>B</span><span>C</span><span>D</span><span>E</span><span>F</span><span>G</span><span>H</span><span>I</span><span>J</span><span>K</span><span>L</span><span>M</span><span>N</span><span>O</span><span>P</span><span>Q</span><span>R</span><span>S</span><span>T</span><span>U</span><span>V</span><span>W</span><span>X</span><span>Y</span><span>Z</span><span>a</span><span>b</span><span>c</span><span>d</span><span>e</span><span>f</span><span>g</span><span>h</span><span>i</span><span>j</span><span>k</span><span>l</span><span>m</span><span>n</span><span>o</span><span>p</span><span>q</span><span>r</span><span>s</span><span>t</span><span>u</span><span>v</span><span>w</span><span>x</span><span>y</span><span>z</span><span>1</span><span>2</span><span>3</span><span>4</span><span>5</span><span>6</span><span>7</span><span>8</span><span>9</span><span>0</span><span>+</span><span>=</span><span>!</span><span>@</span><span>#</span><span>%</span><span>$</span><span>%</span><span>^</span><span>&amp;</span>
             </div>
           </div>
+          <div class="font-family-grid">
```
argyleink, line 2153: really need a component model don't we! there's astro components and svelte components in the new V2 docs.. will make all this much easier to manage. 

i think for this docsite (v1) it's totally fine to have all this repeated markup.

src/props.fonts.css
```diff
@@ -2,6 +2,21 @@
   --font-sans: system-ui,-apple-system,Segoe UI,Roboto,Ubuntu,Cantarell,Noto Sans,sans-serif;
   --font-serif: ui-serif,serif;
   --font-mono: Dank Mono,Operator Mono,Inconsolata,Fira Mono,ui-monospace,SF Mono,Monaco,Droid Sans Mono,Source Code Pro,monospace;
+  --font-system-ui: system-ui, sans-serif;
+  --font-transitional: Charter, 'Bitstream Charter', 'Sitka Text', Cambria, serif;
+  --font-old-style: 'Iowan Old Style', 'Palatino Linotype', 'URW Palladio L', P052, serif;
+  --font-humanist: Seravek, 'Gill Sans Nova', Ubuntu, Calibri, 'DejaVu Sans', source-sans-pro, sans-serif;
```
argyleink, line 8: same comment here about quotes around the font names.

src/props.fonts.js
```diff
@@ -1,25 +1,55 @@
...
-  '--font-lineheight-3': '1.5',
-  '--font-lineheight-4': '1.75',
-  '--font-lineheight-5': '2',
+  "--font-lineheight-00": ".95",
+  "--font-lineheight-0": "1.1",
+  "--font-lineheight-1": "1.25",
+  "--font-lineheight-2": "1.375",
+  "--font-lineheight-3": "1.5",
+  "--font-lineheight-4": "1.75",
+  "--font-lineheight-5": "2",
```
argyleink, line 52: looks like a formatter changed all the quotes to it's preference and made it into the PR. can you trim these quote change commits from the PR please? 🙂

argyleink (changes requested):
overall the PR looks great! i've got a couple open questions to consider:
1. are these fonts better as an optional import rather than being part of the default bundle?
2. is the typography documentation section overwhelming now with all the typography example blocks? we could consider using the `<details>` element OR make a switcher that lets people choose a font prop from a dropdown, which updates an example or all the typography examples
3. how can we give @danklammer and Modern Font Stacks more cred for the values?

i'll pose it to the group in discord and see how folks feel about it

thanks for this, i'm excited about it 🙂

docsite/index.html
Jothsa: According to MDN, they're recommended but not required 🤷🏼‍♂️

src/props.fonts.js
Jothsa: Sorry I missed that 😅

Jothsa:
Good point about giving them more credit. I think it would also be nice to have some information explaining why people would want to use these over an external font. Maybe in a paragraph after the "Font Families" heading?

docsite/index.html
```diff
@@ -2114,6 +2114,22 @@ <h4>Font Families</h4>
           --font-sans: system-ui, -apple-system, Segoe UI, Roboto, Ubuntu, Cantarell, Noto Sans, sans-serif;
           --font-serif: ui-serif, serif;
           --font-mono: Dank Mono, Operator Mono, Inconsolata, Fira Mono, ui-monospace, SF Mono, Monaco, Droid Sans Mono, Source Code Pro, monospace;
```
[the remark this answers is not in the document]
Jothsa: I agree that these could be removed if modern font stacks are added.

Jothsa:
I've got finals this week and next week so I won't be able to do much here until after that

docsite/index.html
```diff
@@ -2114,6 +2114,22 @@ <h4>Font Families</h4>
           --font-sans: system-ui, -apple-system, Segoe UI, Roboto, Ubuntu, Cantarell, Noto Sans, sans-serif;
```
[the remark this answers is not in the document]
argyleink: I wrote a reply here, but after thinking about it, it's all legacy and designer preference. The point for me generally, is to give the user a font they're used to, not one we think looks better. That's what we load fonts and change headers for if we want, but not body copy.

I could def see a version of Open Props updated to shed that legacy preference noise. 
Opened https://github.com/argyleink/open-props/issues/500

docsite/index.html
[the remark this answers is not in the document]
argyleink: let's remove it! I agree. if you include it in this PR, go ahead and indicate in the commit it fixes https://github.com/argyleink/open-props/issues/500 👍🏻

argyleink:
> I've got finals this week and next week so I won't be able to do much here until after that

hope finals went well! let us know if you need help closing this out. it's a sweet addition.

Jothsa:
They did (still waiting for some results though 😶)! Thanks! So sorry for the delay— life’s been coming at me and my ADHD has had me hyperfixating on another project (built with open-props) lol. I’ll get back to this pr soon!

src/props.fonts.js
Jothsa: Should be fixed

docsite/index.html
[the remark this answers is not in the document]
Jothsa: Removed

Jothsa:
Alright, I think that's everything except if you want them split into a separate import. Do you want to keep `--font-system-ui`? It's from modern font stacks, but I think it's unnecessary. It is set to `system-ui, sans-serif`

argyleink:
> Alright, I think that's everything except if you want them split into a separate import. Do you want to keep `--font-system-ui`? It's from modern font stacks, but I think it's unnecessary. It is set to `system-ui, sans-serif`

rad ty! i'm just returning from conferencing and ready to review this. I do think it's a good idea to include that prop as it's often missed by folks and ChromeOS for example doesn't yet support `system-ui` while it does `sans-serif`. healthy fallback habit. Thanks for asking!

I'll give this a review now 👍🏻

src/extra/normalize.src.css
```diff
@@ -15,7 +15,7 @@
 
 :where(html) {
   block-size: 100%;
-  font-family: var(--font-sans);
+  font-family: system-ui;
```
argyleink, line 18:
```suggestion
  font-family: var(--font-system-ui);
```
jit-props will make sure the only one in the bundle is this value

argyleink (approved):
looks great! i made 1 little suggestion which i think i can just inject/accept and this is g2g. appreciate the work!

src/props.fonts.css
```diff
@@ -1,7 +1,19 @@
 :where(html) {
-  --font-sans: system-ui,-apple-system,Segoe UI,Roboto,Ubuntu,Cantarell,Noto Sans,sans-serif;
-  --font-serif: ui-serif,serif;
-  --font-mono: Dank Mono,Operator Mono,Inconsolata,Fira Mono,ui-monospace,SF Mono,Monaco,Droid Sans Mono,Source Code Pro,monospace;
```
argyleink, line 4: actually, 1 more comment, as changing these names means i need to make a breaking change release. alternatively, we could keep these same names or just keep the original props, that way it's a minor version bump and no one's apps break.

which is better?
1. keep the previous props in additional to the new ones
1. create vars that map the old names to the most relevant new ones
1. keep the previous props and make the new ones an additional/optional import

i'll toss this into the discord too, see what folks think!

aslak01: I think option 1 or 2 is best. I kinda prefer 2 but I appreciate the hesitation to have breaking changes so I'd not complain at either. In my opinion these are small enough that I don't think we need to hide them behind a flag, and I love the idea of having the prototyping power of these well thought out props easily available. Also most people who care about the css weight should probably be using jit.

Jothsa: My vote is to map the old props them to the new ones and remove them in the next major version. Maybe add a deprecation notice to the doc site?

mobalti:
I think we should consider the overall approach for introducing new props in the library:

- Should we add them to the main import, as we did with `Static Sizes` and `Drawn Borders`? (While they are great, I'm still not convinced they should be in the main bundle.)
- Or should we keep everything as separate imports from now on?

SanderElias:
> I think we should consider the overall approach for introducing new props in the library:
> 
> * Should we add them to the main import, as we did with `Static Sizes` and `Drawn Borders`? (While they are great, I'm still not convinced they should be in the main bundle.)
> * Or should we keep everything as separate imports from now on?

When the added props are entirely new for the current release, they should be imported separately. I fully agree with this.
However, as those in this PR are touching existing stuff, it must be in the main. Otherwise, it would be breaking.
And then, as proposed, fix this in the next version. 
I suggest proceeding with the current plan. 
For completeness, that is:
  - map the old props to the new
  - add all into the main import
  - add deprecation notice in docs (as needed?)
  - move everything into its self-contained import in the next version.

SanderElias (approved):

Jothsa:
I'm added the old values and set this to the new ones. Will postcss-jit-props be able to handle that correctly? Do you think a deprecation notice is needed?

src/props.fonts.css
```diff
@@ -1,7 +1,22 @@
...
+  --font-system-ui: system-ui, sans-serif;
+  --font-transitional: Charter, Bitstream Charter, Sitka Text, Cambria, serif;
```
argyleink, line 3:
```suggestion
  --font-transitional: ui-serif, Charter, Bitstream Charter, Sitka Text, Cambria, serif;
```
this will aid in less side effects in the update

src/props.fonts.css
```diff
@@ -1,7 +1,22 @@
...
+  --font-old-style: Iowan Old Style, Palatino Linotype, URW Palladio L, P052, serif;
+  --font-humanist: Seravek, Gill Sans Nova, Ubuntu, Calibri, DejaVu Sans, source-sans-pro, sans-serif;
+  --font-geometric-humanist: Avenir, Montserrat, Corbel, URW Gothic, source-sans-pro, sans-serif;
+  --font-classical-humanist: Optima, Candara, Noto Sans, source-sans-pro, sans-serif;
+  --font-neo-grotesque: Inter, Roboto, Helvetica Neue, Arial Nova, Nimbus Sans, Arial, sans-serif;
+  --font-monospace-slab-serif: Nimbus Mono PS, Courier New, monospace;
+  --font-monospace-code: ui-monospace, Cascadia Code, Source Code Pro, Menlo, Consolas, DejaVu Sans Mono, monospace;
```
argyleink, line 10:
```suggestion
  --font-monospace-code: Dank Mono,Operator Mono,Inconsolata,Fira Mono,ui-monospace,SF Mono,Monaco,Droid Sans Mono,Source Code Pro,Cascadia Code,Menlo,Consolas,DejaVu Sans Mono,monospace;
```
would prefer we keep the current monospace font stack instead of fully adopt the one from modern font stacks, however I did leave a few from it in here.

src/props.fonts.js
```diff
@@ -1,7 +1,35 @@
...
+    'Rockwell, Rockwell Nova, Roboto Slab, DejaVu Serif, Sitka Small, serif',
+  '--font-antique':
+    'Superclarendon, Bookman Old Style, URW Bookman, URW Bookman L, Georgia Pro, Georgia, serif',
+  '--font-didone':
+    'Didot, Bodoni MT, Noto Serif Display, URW Palladio L, P052, Sylfaen, serif',
+  '--font-handwritten':
+    'Segoe Print, Bradley Hand, Chilanka, TSCu_Comic, casual, cursive',
+  '--font-sans': 'var(--font-system-ui)',
+  '--font-serif': 'var(--font-transitional)',
+  '--font-mono': 'var(--font-monospace-code)',
```
argyleink, line 32: cool, this totally works

argyleink (changes requested):
thanks for adding those props that map the old naming to the new one!

I've made 2 small change requests, and given those look good to you too, I think this PR will be ready to merge!

src/props.fonts.css
Jothsa: I definitely understand not wanting side effects, but I feel that this change loses the benefit of this font stack over just using ui-serif. It's my understanding that transitional fonts are a specific type of serif font. Since ui-serif is first, it will likely always be chosen (where supported). From my perspective the other font families might as well not be there. Unless there's something I'm missing, I'm not a font expert 😅

argyleink: just looked at the support for `ui-serif` and it's only Safari, so sounds like the only devices that will get it will be iPhones. it'll be the same as `ui-rounded` in the `rounded-sans` set.

Jothsa: Alright.  I suppose it's ok, since if you're using this font stack you don't care about the exact font.  it would bug me if this stack could choose a font that is not a transitional font (since not all serif fonts are transitional fonts) when a transitional font was available to the user

argyleink: that seems fair,  but we're also doing this `--font-serif: var(--font-transitional);` so I guess we'd need a new set of values for font-serif that excludes the transitional serif fonts? that seems weird tho? looking at the set of fonts we have, that seems like the best one to drop it into yeah?

Jothsa: Could we do `--font-serif: ui-serif, serif;`? Since it'll be removed in the next major update I don't see the harm in the duplication

argyleink: works for me!

Jothsa: Done!

Jothsa:
I think the variables should be good to go. I have 2 quick questions. Do you think we need a deprecation notice for the old values? Do you think we should note that
`--monospace-code` is different in open-props than in modern font stacks? I think it's ok to quietly remove the old font values. I also think it's ok to not add the note, but thought I'd run it by you. Thanks for all the help!

argyleink:
> Do you think we need a deprecation notice for the old values? 

I've [made a new issue](https://github.com/argyleink/open-props/issues/505) to hold this task, I'll try and make sure it's part of OPv2 👍🏻 

<br>

> Do you think we should note that `--monospace-code` is different in open-props than in modern font stacks?

Seems like a fair small note to add to the docs. Use a `blockquote` like this? 
Screenshot 2024-07-02 at 9 30 58 AM

argyleink (approved):
hazzah! thanks for all this and stickin it out!!

Jothsa:
No problem! Thanks for your support and time!

PR #528, opened 2024-12-03
Add negative sizes as a prop pack

felix-bohlin:
Not sure if I nailed the naming of stuff, you might want to take a look at that. 

Worth noting also is that I reversed the values in each clamp and switched to logical units, ie:
```css
--size-fluid-1: clamp(0.5rem, 1vw, 1rem); /* positive */
--size-fluid-1-: clamp(-1rem, -1vi, -0.5rem); /* negative */

```

### DIFF
--- a/css/packs/negative-sizes/fluid.css
+++ b/css/packs/negative-sizes/fluid.css
@@ -1,12 +1,6 @@
 :where(:root) {
-	--size-fluid-1-: clamp(-1rem, -1vi, -0.5rem);
-	--size-fluid-2-: clamp(-1.5rem, -2vi, -1rem);
-	--size-fluid-3-: clamp(-2rem, -3vi, -1.5rem);
-	--size-fluid-4-: clamp(-3rem, -4vi, -2rem);
-	--size-fluid-5-: clamp(-5rem, -5vi, -4rem);
-	--size-fluid-6-: clamp(-7.5rem, -7vi, -5rem);
-	--size-fluid-7-: clamp(-10rem, -10vi, -7.5rem);
-	--size-fluid-8-: clamp(-15rem, -20vi, -10rem);
-	--size-fluid-9-: clamp(-20rem, -30vi, -15rem);
-	--size-fluid-10-: clamp(-30rem, -40vi, -20rem);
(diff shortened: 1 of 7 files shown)


argyleink (commented):
what's the use case for these? do they actually produce usable values?

curious if we should instead just

```css
--size-fluid-1-: calc(var(--size-fluid-1) * -1);
```

this produces the inverted value too?

felix-bohlin:
The point of negative fluid sizes specifically or all of them?

There's value in negative versions of size and relative. Fluid is probably a more contrived situation but my thought process there was that
a) my lack of imagination how these primitives should be used shouldn't stop others from doing it 
b) it wouldn't "burden" the core lib since it would be a prop pack  

With that said, there's no rush here. More code is more technical debt and once you release something it's hard to take it back. 

Regarding your proposition with calc though - it looks so much cleaner. Yes.

argyleink:
> a) my lack of imagination how these primitives should be used shouldn't stop others from doing it
b) it wouldn't "burden" the core lib since it would be a prop pack

yep, very true. however, I do need at least one or two use cases in order to verify the negative value strategy we go with, is actually useful.

like here, you show not just inverting values, but inverting the parameters:

```css
--size-fluid-1: clamp(0.5rem, 1vw, 1rem); /* positive */
--size-fluid-1-: clamp(-1rem, -1vi, -0.5rem); /* negative */
```

that's interesting!

and here I suggest inverting the result of the value it's a compliment for:

```css
--size-fluid-1-: calc(var(--size-fluid-1) * -1);
```

🤷🏻 

I don't think these two strategies produce the same values in all situations, and I don't know which one is right. I think we need to know what we're solving for, to make sure these vars behave like we want them to.

best idea I can come up with is some negative margin to offset the padding or margin of a container?

felix-bohlin:
Regardless of how interesting it could be to invert all values I think your inversion is more what you'd expect from a negative size-fluid - it's way more reasonable. The use-cases won't differ much between the two approaches as far as I can tell.

[This is a demo](https://codepen.io/Flerox/pen/vEBNNwE?editors=1100) I made two weeks ago and it's all negative margins and pull quotes and break out stuff. Can't think of cooler stuff than that 😅

PR #531, opened 2024-12-10
Add interpolate-size to normalize

felix-bohlin:
If there are some gotchas with having this on the `:root`, let me know. Otherwise thought this would be neat to include out of the box. A downside may be that it doesn't have enough of a graceful fallback, idk.


argyleink (approved):
fixes #512

PR #535, opened 2024-12-21
chore: add field-sizing to select and textarea

kaf-lamed-beyt:


src/extra/normalize.src.css
```diff
@@ -130,14 +130,15 @@
...
+  field-sizing: content;
 }
 
 :where(textarea, select, input:not([type="button"],[type="submit"],[type="reset"])) {
   background-color: var(--surface-2);
   border-radius: var(--radius-2);
 }
 
-:where(textarea) { resize: block }
+:where(textarea) { resize: block; field-sizing: content }
```
argyleink, line 141:
```suggestion
:where(textarea) { 
  resize: block;
  field-sizing: content;
  min-inline-size: var(--size-content-1);
  min-block-size: 2lh;
}
```

this ensures the textarea doesnt simply collapse once field-sizing: content is applied.

argyleink (changes requested):
i've got one change to throw in there before field-sizing can merge, made it as a suggestion here 👍🏻

kaf-lamed-beyt:
> i've got one change to throw in there before field-sizing can merge, made it as a suggestion here 👍🏻

wonderful!

PR #538, opened 2024-12-30
Add text-decoration: none to .btn links for better styling

mobalti:
This PR adds `text-decoration: none` to the `.btn` class scoped specifically to links. This ensures that links styled as buttons have consistent appearance without requiring additional author styling.
The selector used is `:where(.btn:is(a))`, which emphasizes `.btn` as the primary styling hook.

### DIFF
--- a/src/extra/buttons.css
+++ b/src/extra/buttons.css
@@ -101,6 +101,10 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


argyleink (approved):
ty!

Issue #304, opened 2022-10-31
document light/dark switch imports


argyleink:

kaf-lamed-beyt:
@argyleink, I can see that you have something similar documented here.

![image](https://github.com/user-attachments/assets/8ae4badc-f3f1-41a9-bf19-9c9a66d8b023)

Is this correct?

argyleink:
Similar, but with a switch you have a way to pick a theme and persist it across page loads

kaf-lamed-beyt:
oh! How do you want this to be documented @argyleink? Is this the section I'll be including this info? and am I to mention using a switch? 
trying to get more context

argyleink:
it's more about documenting these [switch imports](https://github.com/argyleink/open-props/blob/main/src/extra/theme.light.switch.css) so it's clear what they are and why you'd need them.

[Demo 1](https://codepen.io/argyleink/pen/RwEWRwR)  
[Demo 2](https://codepen.io/argyleink/pen/JjqYdgr)

**what and why?** they're new scopes for light or dark modes, allowing authors to either assign a theme to a page or a section. this is most commonly used with a theme switch on a side, like a moon or sun toggle, that can assign a theme to the page via JavaScript and local storage. the default of normalize is "auto follow the system theme" and these switches allow "always be dark mode" type of user assignment.

kaf-lamed-beyt:
Oh okay. This makes sense now. 

Where on the page would you prefer this to be?

argyleink:
<img width="444" alt="Screenshot 2024-12-30 at 9 00 02 AM" src="https://github.com/user-attachments/assets/cfcf6e33-893b-4759-a814-adf7a6667c60" />

let's make it `Color Theming 201: usage with client side switches`

kaf-lamed-beyt:
aha! alright. I'll get on this

PR #540, opened 2025-01-01
chore: document theme switch imports

kaf-lamed-beyt:
#304 

How does this look @argyleink?

image


argyleink (changes requested):
this looks great so far! thanks a bunch ❤️ I can see the effort put in.

I had enough edits that I made a PR to your fork. Mind accepting those, then this branch/PR will update and we can see if there are last minute edits to make?

kaf-lamed-beyt:
oh alright. thanks for the feedback. I'll take a look at the PR now

src/extra/normalize.src.css
```diff
@@ -138,7 +138,7 @@
   border-radius: var(--radius-2);
 }
 
-:where(textarea) { 
+:where(textarea) {
```
argyleink, line 141: this is a strange commit? can you remove it so the PR is just on the index.html file?

kaf-lamed-beyt: huh! That must have been from prettier. I must have pressed `ctrl + shift + i` to format the file manually. My bad.

kaf-lamed-beyt: Looking at it now... I don't know how to actually update it without creating a commit in that file.

image

Although I pulled from my remote now and I can see that index.html has untracked changes. Perhaps pushing that change may update `normalize.src.css`? 

image


What do you think?

argyleink: i'll squash this work in, so the history isn't that big of a deal. if you just push an update to the branch that removes this commit, that should be good enough. i'm not a stickler about the file history, and yeah, we'll be squashing so shouldnt be an issue.

kaf-lamed-beyt: oh okay. i'll dod this now. 

got stuck on a couple of things, hence the reason I'm replying now 🙏🏼

PR #542, opened 2025-01-28
Fix: Ensure consistent fonts and correct usage of --font-monospace-code

mobalti:
- Fixed inconsistent formatting in `props.fonts.css`, adding proper spacing to the `--font-monospace-code` prop value. 
- Corrected an issue in the doc where `font-family: var(--font-monospace);` was mistakenly used instead of `font-family: var(--font-monospace-code);`


argyleink (approved):
looks great thanks!

the package-lock is sneaking in, but should be ok

PR #548, opened 2025-02-11
Added flinch to durations.css

mneumegen:
I'm using these to time transitions and regularly use something in the 300ms range.


argyleink (approved):
love it!

PR #553, opened 2025-03-20
Optimizing the mask gradients

Afif13:
I found that we can still reduce the number of gradients for the zig-zag horizontal and vertical version.


argyleink (approved):
good call, we should done this earlier!

Issue #397, opened 2023-08-24
gradients should auto upgrade to `oklab` and HDR


argyleink:

mobalti:
I think for the colors it's okay, but for the gradients, we have to wait for Firefox support.
https://developer.mozilla.org/en-US/docs/Web/CSS/hue-interpolation-method

mobalti:
In the meantime, we also need more gradients.

argyleink:
this could be something like:

```css
:root {
  --space: ;
  --gradient-1: linear-gradient(to right var(--space), red, blue);
  --gradient-2: …;
  --gradient-3: …;
}

@supports (background: linear-gradient(to right in oklab, #000, #fff)) {
  :root {
    --space: in oklab;
  }
}
```

PR #554, opened 2025-03-30
feat: auto-upgrade gradients to oklab and HDR (#397)

mobalti:
- Upgraded all gradients to use Oklab for better color transitions.
- Preserved original colors while transitioning to Oklab.


argyleink (changes requested):
love that you kicked this off!

so the gradients continue to work in older browsers and don't break anyone's site, let's go this route

```css
:root {
  --space: ;
  --gradient-1: linear-gradient(to right var(--space), red, blue);
  --gradient-2: …;
  --gradient-3: …;
}

@supports (background: linear-gradient(to right in oklab, #000, #fff)) {
  :root {
    --space: in oklab;
  }
}
```

would also let someone switch it to `in oklch` also, which is what Tailwind 4 went with and tends to be more "vibrant" but possibly more surprising too.

mobalti:
- Ensures better support for older browsers.
- Adds support for customizable color spaces in gradients.   
- All tests passed, and everything works correctly when tested locally with Vite.
- However, the docsite’s PostCSS version doesn’t handle `--space: ;` properly, causing an error (details shared earlier).
- **PostCSS should be updated before merging** to prevent breaking the docsite.

src/props.gradients.css
```diff
@@ -1,62 +1,63 @@
 :where(html) {
-  --gradient-1: linear-gradient(to bottom right,#1f005c, #5b0060, #870160, #ac255e, #ca485c, #e16b5c, #f39060, #ffb56b);
-  --gradient-2: linear-gradient(to bottom right,#48005c, #8300e2, #a269ff);
+  --space: ;
```
argyleink, line 2:
```suggestion
  --gradient-space: ;
```

sorry for asking you to rename a bunch of stuff… but space seemed super generic, and if we prefix it with --gradient- then it's much clearer what part it plays.

test/basic.test.cjs
```diff
@@ -4,7 +4,7 @@ const OpenProps = require('../dist/open-props.cjs')
 const OPtokens = require('../open-props.tokens.json')
 
 test('Should have an all included import', t => {
-  t.is(Object.keys(OpenProps).length, 1734)
+  t.is(Object.keys(OpenProps).length, 1736)
```
argyleink, line 7: excellent, ty!

argyleink (changes requested):
sorry for the delay on this review, but I'm here now!

src/props.gradients.css
mobalti: Totally agree, especially with HDR colors, it's important to be specific.

mobalti:
This sets the foundation for the upcoming gradients. Would love to see many of the defaults from `gradients.style` eventually available as props!

argyleink (approved):
looks great! tyvm!!

argyleink:
these tests… i've kicked them 4 times now, they're randomly failing. will merge once they're complete

argyleink:
oh, does look like a test issue https://github.com/argyleink/open-props/actions/runs/14283660499/job/40118537536#step:6:213 with the new prop.

argyleink:
> Updating postcss-preset-env resolves the issue:

are there any side effects from the update? down to bump that package in this PR and we'll see if all the tests pass? we already know the docsite still looks the same after the update, so i dont anticipate any issues. but it would be good to bump the version with the PR that needs the update, keep the history 👍🏻

argyleink:
releases in `v1.7.15` 🎉   
ty for this, great work!

Issue #557, opened 2025-06-01
Using container as custom media queries


AlessioCoser:
Do you see any problem handling custom media queries without the need of postcss preprocessor?

I thought something like this could do:

```
:root {
    --is-below-xxs: false;
    --is-below-xs: false;
    --is-below-sm: false;
    --is-below-md: false;
    --is-below-lg: false;
    --is-below-xl: false;
    --is-below-xxl: false;

    @media screen and (max-width: 240px) {
        --is-below-xxs: true;
    }
    @media screen and (max-width: 360px) {
        --is-below-xs: true;
    }
    @media screen and (max-width: 480px) {
        --is-below-sm: true;
    }
    @media screen and (max-width: 768px) {
        --is-below-md: true;
    }
    @media screen and (max-width: 1024px) {
        --is-below-lg: true;
    }
    @media screen and (max-width: 1440px) {
        --is-below-xl: true;
    }
    @media screen and (max-width: 1920px) {
        --is-below-xxl: true;
    }
}
```

Usage:
```
@container style(--is-below-xxs: true) {
    /* ... */
}
```

I've checked the browser compatibility, and it seems fine: https://caniuse.com/?search=%40container

What do you think? I can try to make a pull request if you think it would be a good fit for the project.

argyleink:
hi @AlessioCoser!

fun idea, reminds me of https://github.com/propjockey/css-media-vars except instead of using tricky dry switching you lean into style queries. worth noting here, that [style queries](https://caniuse.com/css-container-queries-style), which do have different support than container queries, have less support atm. good news is, Open Props is interested in standards and doesnt focus on browser support, so this'll eventually be a non-issue.

tldr;
i think this is cool, and yeah, does have a good fit here and I'm down to help you land a PR into this repo so others can use your cool work 🙂 

here's the [media queries usable from Open Props today](https://github.com/argyleink/open-props/blob/main/src/props.media.css), they lean into the @custom-media draft standard (which needs a preprocessor like you mention), but does show a bunch of media queries that folks would likely want a variable for.

PR #558, opened 2025-06-02
Resolves #447 - Use @container style as custom media query

AlessioCoser:


AlessioCoser:
@argyleink it is not ready yet, but i would like a feedback if i'm on the right track

AlessioCoser:
https://github.com/argyleink/open-props/issues/557

AlessioCoser:
I kinda followed the animation file more or less, and generated accordingly the required @media.

@argyleink this should be a working version. Tell me if you see something wrong

src/index.js
```diff
@@ -2,6 +2,7 @@ import Animations from './props.animations.js'
 import Sizes from './props.sizes.js'
 import Colors from './props.colors.js'
 import ColorsHSL from './props.colors-hsl.js'
+import Container from "./props.container.js";
```
argyleink, line 5:
```suggestion
import Container from './props.container.js'
```
match style of others

argyleink (changes requested):
found one minor syntax nit, no biggie

the last task afaict, is to include these new props in the JS object output too, for the css-in-js users. this just means opening up `build/to-object.js` and adding the container values to the lists. 

then update tests: expected as this PR is adding a lot of new props. you can run the tests locally to see the new total, then update that value in [the test](https://github.com/argyleink/open-props/blob/main/test/basic.test.cjs#L7), and then this PR should be good to go 👍🏻

Issue #526, opened 2024-11-29
Include Penner Easing equations in easing.css


gordonbrander:
[Robert Penner's easing equations](http://robertpenner.com/scripts/easing_equations.txt) are a battle-tested set of Bézier curves for plush animations:

- Quadratic
- Cubic
- Quartic
- Quintic
- Sine
- Exponential
- Circular

E.g. `easeInOutSine` `easeOutExpo`, etc.

These date back to the Flash days and make up a common default "kit" for easing (with the addition of springs).

Two CSS implementations:
- https://matthewlein.com/tools/ceaser
- https://gist.github.com/amadeus/983364

argyleink:
these were in the original set of easings for Open Props! they were leveraging this postcss plugin https://www.npmjs.com/package/postcss-easings

and i agree, these are battle tested and often the first set that folks get introduced to when exploring the potential curves… i just always found their names confusing as hell to remember while also never really loving the results. BUT, just because I dont love them, doesnt mean they shouldnt be a part of this CSS variables collection 👍🏻 

+1 to adding them back so the collection is the classic "kit" plus moarrrr

PR #559, opened 2025-07-04
Add Robert Penner's easing functions (#526)

maltebaer:
As discussed in #526, I have added seven easing functions based on Robert Penner's equations.

I took the implementation for approximating Penner's equations using cubic Bézier curves from https://github.com/matthewlein/Ceaser/blob/master/developer/ceaser-easings.js.

### DIFF
--- a/docsite/index.html
+++ b/docsite/index.html
@@ -2520,7 +2520,7 @@
     <section id="easing">
       <header>
         <h2>Easing</h2>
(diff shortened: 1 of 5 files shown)


argyleink (changes requested):
Thanks for the PR!

Tested locally, everything looks good. 

The tests are easy enough to fix:
- https://github.com/argyleink/open-props/blob/main/test/basic.test.cjs#L7 is `1778` now
- https://github.com/argyleink/open-props/blob/main/test/basic.test.cjs#L27 is `484`

